PR #6, opened 2024-08-02
Feat: replace mixHash to difficulty

jennwiederholen:
Deprecate IstanbulDigest, which is constant value of mixHash field that represents certain block is qbft block, due to following reason : 

1. difficulty field also represents constant value for qbft logic. we can replace with it
2. mixHash field is currently used for randao value in ethereum. (eip-4399). wemix may have to use that field too.


consensus/qbft/backend/engine_test.go
```diff
@@ -265,14 +265,15 @@ func TestVerifyHeader(t *testing.T) {
 		t.Errorf("error mismatch: have %v, want %v", err, qbftcommon.ErrInvalidExtraDataFormat)
 	}
 
+	// TODO This test does not need anymore as we don't fix mixHash field to IstanbulDigest
```
kai-yu-wm, line 268: I think it would be better to keep the test code and refer to the link below for the related code.
https://github.com/wemixarchive/go-wemix-qbft/blob/feat/deprecate-istanbul-digest/consensus/clique/clique.go#L284

core/types/istanbul_test.go
```diff
@@ -32,10 +32,10 @@ import (
 func TestHeaderHash(t *testing.T) {
 	// 0xcefefd3ade63a5955bca4562ed840b67f39e74df217f7e5f7241a6e9552cca70
```
kai-yu-wm:
```suggestion
	// 0x06f58ebb96f233516c4a0eeedad007e8b8e0965b24c88d7407eb08a6ee7bc4ff
```

consensus/qbft/backend/engine_test.go
jennwiederholen: to make the test code make sense, i guess i need to add verifying  mixHash in `verifyHeader()` function which i currently deleted.
```
func (e *Engine) verifyHeader(chain consensus.ChainHeaderReader, header *types.Header, parents []*types.Header, validators qbft.ValidatorSet) error {
	...
	// Ensure that the mix digest is zero as we don't have fork protection currently
	//if header.MixDigest != types.IstanbulDigest {
	//	return qbftcommon.ErrInvalidMixDigest
	//}
	...
```

to use  mixHash as random value field, we don't want  mixHash field  to be zero like `clique` so.. what would be the good way to validate it?

kai-yu-wm: I checked the code incorrectly. when random is added, reconsider it.

kai-yu-wm (approved):

PR #21, opened 2024-10-08
fix: qbft test failures

egonspace:
make qbft test to pass
(not all tests)


consensus/qbft/engine/engine.go
```diff
@@ -319,7 +319,6 @@ func (e *Engine) VerifySeal(chain consensus.ChainHeaderReader, header *types.Hea
 }
 
 func (e *Engine) Prepare(chain consensus.ChainHeaderReader, header *types.Header, validators qbft.ValidatorSet) error {
-	header.Coinbase = common.Address{}
```
egonspace, line 322: This is meaningless.
Before calling engine.Prepare, miner module sets `Coinbase` and engine reset `Coinbase` in `Seal`.

PR #24, opened 2024-10-10
add prepare msg to block

paul-ahn-wm:


consensus/qbft/core/prepare.go
```diff
@@ -36,7 +37,19 @@ func (c *Core) broadcastPrepare() {
...
-	prepare := qbftmessage.NewPrepare(sub.View.Sequence, sub.View.Round, sub.Digest)
+
+	var header *types.Header
+	if block, ok := c.current.Proposal().(*types.Block); ok {
+		header = block.Header()
+	}
+	// Create Prepare Seal
+	prepareSeal, err := c.backend.SignWithoutHashing(PrepareCommittedSeal(header, uint32(c.currentView().Round.Uint64())))
+	if err != nil {
+		logger.Error("QBFT: failed to create COMMIT seal", "sub", sub, "err", err)
```
jennwiederholen: COMMIT seal -> PREPARE seal

consensus/qbft/engine/engine.go
```diff
@@ -247,6 +278,85 @@ func (e *Engine) verifySigner(chain consensus.ChainHeaderReader, header *types.H
...
+
+	if len(prevPreparedSeal) == 0 {
+		// prevPreparedSeal validation for monblanc block is skipped because it's empty
+		if !(chain.Config().MontBlancBlock.Cmp(header.Number) == 0) {
+			return qbftcommon.ErrEmptyPrevPreparedSeals
+		}
+	} else {
+		//check whether prevPrepared seals are generated by prevValidators
+		var prevPreparers []common.Address
+		prevHeader := chain.GetHeaderByNumber(number - 1)
```
jennwiederholen: Need to get prevHeader  from parents rather than canonical chain. 
node uses this function when syncing chain and while syncing, node validates header"s" in bulk way so prev Header may not exist in chain.

paul-ahn-wm: I have implemented it the same way as the existing verifyCommittedSeals function, but if there have been changes, I will rewrite it the same way after merging.

consensus/qbft/engine/engine.go
```diff
@@ -247,6 +278,85 @@ func (e *Engine) verifySigner(chain consensus.ChainHeaderReader, header *types.H
...
+	extra, err := types.ExtractQBFTExtra(header)
+	if err != nil {
+		return err
+	}
+
+	prevPreparedSeal := extra.PrevPreparedSeal
+
+	if len(prevPreparedSeal) == 0 {
+		// prevPreparedSeal validation for monblanc block is skipped because it's empty
+		if !(chain.Config().MontBlancBlock.Cmp(header.Number) == 0) {
```
jed-wemade, line 299:
```go
if !chain.Config().IsMontBlanc(header.Number) {
```

egonspace:
I agree with the overall implementation direction.

consensus/qbft/engine/engine.go
jennwiederholen: actually, validation exception should be applied only to montblac hardfork.  i think comparing exact block number is appropriate.

jennwiederholen (approved):

consensus/qbft/engine/engine.go
```diff
@@ -493,7 +615,29 @@ func (e *Engine) ExtractGenesisValidators(header *types.Header) ([]common.Addres
...
 }
 
-func (e *Engine) Signers(header *types.Header) ([]common.Address, error) {
+func (e *Engine) PrepareSigners(header *types.Header) ([]common.Address, error) {
+	extra, err := types.ExtractQBFTExtra(header)
+	if err != nil {
+		return []common.Address{}, err
+	}
+	preparedSeal := extra.PreparedSeal
+	proposalSeal := PrepareCommittedSeal(header, extra.Round)
```
egonspace, line 624: How about to combine the same codes with `CommitSigners`?

PR #33, opened 2024-10-30
fix test case graphql internal rpc

kai-yu-wm:
- enable graphql & internal & rpc test
  - graphql : skip related to Shanghai Fork
  - internal : modify the JSON resul file (add header field and change some values)
  - rpc : add fields to header (fees, rewards, minerNodeId, minerNodeSig)


kai-yu-wm:
rework

PR #36, opened 2024-10-31
feat: add PrevPreparedSeal and PrevCommittedSeal to the block extra

paul-ahn-wm:
* Fields for prevPrepared, prevCommitted seals have been added to qbft extra field.
* Reward distribution codes for preapare and commit messages have been Implemented.
* Relevant test codes have been written.
* Concurrency issues have been fixed.


consensus/qbft/engine/engine.go
```diff
@@ -170,7 +191,7 @@ func (e *Engine) VerifyHeaders(chain consensus.ChainHeaderReader, headers []*typ
 			if errored {
 				err = consensus.ErrUnknownAncestor
 			} else {
-				err = e.verifyHeader(chain, header, headers[:i], validators)
+				err = e.verifyHeader(chain, header, headers[:i], validators, prevValidators) // BUG? 헤더마다 valset 다를텐데 검증 가능한지..? 해당 함수가 사용되는 부분은 없어보임.
```
egonspace: Please remove hangul comment. And this code seems to have no problem.

consensus/qbft/backend/engine.go
```diff
@@ -74,12 +81,23 @@ func (sb *Backend) VerifyHeader(chain consensus.ChainHeaderReader, header *types
...
 	// Assemble the voting snapshot
-	snap, err := sb.snapshot(chain, header.Number.Uint64()-1, header.ParentHash, parents)
-	if err != nil {
+	var snap, prevSnap *Snapshot
+	var err error
+
+	if snap, err = sb.snapshot(chain, header.Number.Uint64()-1, header.ParentHash, parents); err != nil {
+		return err
+	} else if header.Number.Uint64() < 2 {
+		//TODO: related to genesis block's valset
```
egonspace: In case of block 1, I think `snap.ValSet` is enough to verify that block so you can remove `TODO` tag in the comment.

consensus/qbft/backend/engine.go
```diff
@@ -74,12 +81,23 @@ func (sb *Backend) VerifyHeader(chain consensus.ChainHeaderReader, header *types
...
+		return sb.Engine().VerifyHeader(chain, header, parents, snap.ValSet, snap.ValSet)
+	} else if len(parents) < 1 {
+		if prevSnap, err = sb.snapshot(chain, header.Number.Uint64()-2, chain.GetHeaderByNumber(header.Number.Uint64()-2).Hash(), nil); err != nil {
```
egonspace, line 93: You should add the checking code for that `chain.GetHeaderByNumber(header.Number.Uint64()-2)` might be nil.

jennwiederholen: If chain.GetHeaderByNumber(header.Number.Uint64()-2 is nil, it will return error.UnknownAncestor inside snapshot() func. 
So, if we want to handle nil block before snapshot(), would it be better to return another error than UnknowAncestor?

consensus/qbft/backend/engine.go
```diff
@@ -74,12 +81,23 @@ func (sb *Backend) VerifyHeader(chain consensus.ChainHeaderReader, header *types
...
+			return err
+		}
+	} else if prevSnap, err = sb.snapshot(chain, header.Number.Uint64()-2, parents[0].Hash(), parents[:1]); err != nil {
```
egonspace: I think we cannot use parents[0].Hash() in this because `parents` is just a hint for using cached blocks, so we must use the parent of the `header.ParentHash` block in this.

egonspace: On second thought, once the staking gov contract is implemented, there will be no need to use `snapshot()`, so this code will need to be modified later anyway. Therefore, it seems there's no need to change it right now.

jennwiederholen: yes the snapshot logic will be removed when staking gov is added. however before that, i fixed the logic error :) thank you!

consensus/qbft/engine/engine.go
```diff
@@ -260,6 +368,40 @@ func (e *Engine) verifyCommittedSeals(chain consensus.ChainHeaderReader, header
...
+	prevCommittedSeal := extra.PrevCommittedSeal
+
+	// Check whether the prevCommittedSeals are generated by parent block's validators
+	if len(prevCommittedSeal) == 0 {
+		// prevCommittedSeal validation for monblanc block is skipped because it's empty
+		if !(chain.Config().MontBlancBlock.Cmp(header.Number) == 0) {
+			return qbftcommon.ErrEmptyPrevCommittedSeals
+		}
+	} else {
+		var prevCommitters []common.Address
```
egonspace: How about to define a common `verifySealers` function like this?
```
func (e *Engine) verifySealers(sealers []common.Address, validators qbft.ValidatorSet) error {
	validatorsCpy := validators.Copy()

	// Check whether the committed seals are generated by validators
	validSeal := 0

	for _, addr := range sealers {
		if validatorsCpy.RemoveValidator(addr) {
			validSeal++
			continue
		}
		return qbftcommon.ErrInvalidSeals
	}

	// The length of validSeal should be larger than number of faulty node + 1
	if validSeal <= validators.F()*2 {
		return qbftcommon.ErrInvalidSeals
	}

	return nil
}
```

jennwiederholen: great idea. i'll apply it

consensus/qbft/engine/engine.go
```diff
@@ -247,8 +278,85 @@ func (e *Engine) verifySigner(chain consensus.ChainHeaderReader, header *types.H
...
+
+		for _, addr := range prevPreparers {
+			if prevValidatorsCpy.RemoveValidator(addr) {
+				prevValidSeal++
+				continue
+			}
+			return qbftcommon.ErrInvalidPrevPreparedSeals
+		}
+
+		// The length of validSeal should be larger than number of 2*faulty node + 1
```
jed-wemade: Strictly, we should gather more than or equal **n - f** seals.
For example, n = 5, f = 1, 4 seals make quorum, which means 3 seals (= 2f + 1) are not enough.

Similarly for `QuorumSize()` and `handleRoundChangeMsg()`.

egonspace: If `f` would be given and `n` would be derived from `f`, then `2*f+1` is correct for all `f`. But in our code, `n` is given and `f` is derived from `n`, we should use `QuorumSize()`.
`n-f` is a sufficient condition, but it is too excessive as a necessary condition.

paul-ahn-wm: @egonspace
The current `QuorumSize()` function is implemented as `return int(math.Ceil(float64(2*c.valSet.Size()) / 3))`, when n=7, 4 would be sufficient theoretically. but it returns 5, which requires additional unnecessary messages.
 Thus, simply using quorum size is not accurate, and if we want to use quorum size, we need to modify the formula as `return int(math.Ceil(float64(c.valSet.Size() + c.valSet.F() + 1) / 2))`. Of course, once this change is made, it will be necessary to verify that other logic using QuorumSize() functions correctly without issues.
 If we want to implement it in a truly strict manner, then implementing it as described above is correct. However, there is room for consideration as to whether this level of strictness is really necessary.

egonspace: If n=7, then f=2 so we need 5 seals at least not 4. `QuorumSize()` is correct.

egonspace: After a lengthy discussion, we reached the following conclusion: under the following conditions, the existing `QuorumSize()` does not pose a problem.
n=4, f=1
n=5, f=1
n=6, f=1
n=7, f=2

However, in actual BFT assumptions, the theoretical values for f are not as optimistic as above.
n=4, f=1
n=5, f=1.3333...
n=6, f=1.6666...
n=7, f=2
To ensure safety and liveness under these conditions, we need to use n - f. Therefore, we will modify `QuorumSize()` to calculate based on `n - f`.

consensus/qbft/backend/engine_test.go
```diff
@@ -966,7 +966,14 @@ func TestSimulation(t *testing.T) {
...
 				}
-				nodes[i].balance = balance
+				nodes[j].balance = balance
+				break
+			}
+		}
+
+		// wait until the engine is ready to accept the next block
+		for {
+			if engine.core.GetState() == 0 { // StateAcceptRequest
```
egonspace: How about to add `Sleep` for reducing CPU consumption?

event/event.go
```diff
@@ -161,7 +161,7 @@ type TypeMuxSubscription struct {
 }
 
 func newsub(mux *TypeMux) *TypeMuxSubscription {
-	c := make(chan *TypeMuxEvent)
+	c := make(chan *TypeMuxEvent, 10)
```
egonspace: Could you explain why do you use a buffered channel instead of unbuffered channel?

consensus/qbft/backend/engine_test.go
jennwiederholen: Looks better idea. 
@paul-ahn-wm  please add loop term

event/event.go
paul-ahn-wm: In the test case, it was observed that using an unbuffered channel caused concurrency issues as the order of incoming messages was not guaranteed. The code was modified to address this issue. However, after further discussion, it was concluded that maintaining the order of incoming messages was not necessary, so the code was reverted to use an unbuffered channel.

consensus/qbft/backend/engine_test.go
paul-ahn-wm: The test code you commented on has been removed due to concurrency issues.

consensus/qbft/engine/engine.go
```diff
@@ -250,8 +281,78 @@ func (e *Engine) verifySigner(chain consensus.ChainHeaderReader, header *types.H
...
+	extra, err := types.ExtractQBFTExtra(header)
+	if err != nil {
+		return err
+	}
+
+	prevPreparedSeal := extra.PrevPreparedSeal
+
+	if len(prevPreparedSeal) == 0 {
+		// prevPreparedSeal validation for monblanc block is skipped because it's empty
+		if !(chain.Config().MontBlancBlock.Cmp(header.Number) == 0) {
```
egonspace: Shouldn't we skip not only the hard fork block but also the first block created after the genesis block?

jennwiederholen: Yes we should. I'll add the case

consensus/qbft/core/prepare.go
```diff
@@ -36,7 +37,19 @@ func (c *Core) broadcastPrepare() {
...
 	// Create PREPARE message from the current proposal
 	sub := c.current.Subject()
-	prepare := qbftmessage.NewPrepare(sub.View.Sequence, sub.View.Round, sub.Digest)
+
+	var header *types.Header
+	if block, ok := c.current.Proposal().(*types.Block); ok {
+		header = block.Header()
+	}
+	// Create Prepare Seal
+	prepareSeal, err := c.backend.SignWithoutHashing(PrepareCommittedSeal(header, uint32(c.currentView().Round.Uint64())))
```
egonspace: There seems to be a potential issue where the digest of `CommittedSeal` and `PreparedSeal` are identical (header + round), allowing the `PreparedSeal` to be included in the `CommittedSeal` as well.

egonspace: I added a test case for it.
```
func TestReusePreparedSeal(t *testing.T) {
	chain, engine, _ := newBlockChain(1)
	defer engine.Stop()
	normalBlock := makeBlock(chain, engine, chain.Genesis())

	err := engine.VerifyHeader(chain, normalBlock.Header())
	if err != nil {
		t.Errorf("error mismatch: have %v, want nil", err)
	}

	prepareReused := normalBlock.Header()
	extra, _ := types.ExtractQBFTExtra(prepareReused)
	preparedSeal := extra.PreparedSeal
	extra.CommittedSeal = preparedSeal // reuse prepared seal for committed seal
	setExtra(prepareReused, extra)
	err = engine.VerifyHeader(chain, prepareReused)
	if err == nil {
		t.Errorf("prepared seal can be used for committed seal as well")
	} else if !errors.Is(err, qbftcommon.ErrInvalidCommittedSeals) {
		t.Errorf("error is not ErrInvalidCommittedSeals")
	}
}

```

consensus/qbft/engine/engine.go
```diff
@@ -99,8 +120,8 @@ func (e *Engine) VerifyBlockProposal(chain consensus.ChainHeaderReader, block *t
 	}
 
 	// verify the header of proposed block
-	err := e.VerifyHeader(chain, block.Header(), nil, validators)
-	if err == nil || err == qbftcommon.ErrEmptyCommittedSeals {
+	err := e.VerifyHeader(chain, block.Header(), nil, validators, prevValidators)
+	if err == nil || err == qbftcommon.ErrEmptyCommittedSeals || err == qbftcommon.ErrEmptyPreparedSeals {
 		// ignore errEmptyCommittedSeals error because we don't have the committed seals yet
 		return 0, nil
```
egonspace: I think we must not return here and check next conditions. 
There might be a bug where the `prevCommittedSeal` is incorrect but goes unchecked(in the `VerifyHeader()` function). This is because an error would occur during the validation of the `preparedSeal`, causing the `committedSeal` check to be skipped due to that error.

consensus/qbft/core/prepare.go
```diff
@@ -91,11 +104,11 @@ func (c *Core) handlePrepareMsg(prepare *qbftmessage.Prepare) error {
 		return err
 	}
```
egonspace, line 116: I think we need to verify the `prepare.PrepareSeal` here.

egonspace:
`d81db73` commit includes
 - add seal type to seal data to defend a prepare seal reusing
 - verify seal on handling prepare message and commit message
 - fix test data to account for changes in the block hash caused by modifications to the seal

consensus/qbft/core/core.go
```diff
@@ -316,8 +324,24 @@ func (c *Core) checkValidatorSignature(data []byte, sig []byte) (common.Address,
...
+func verifySeal(header *types.Header, round uint32, sealType SealType, seal []byte, sealer common.Address, valSet qbft.ValidatorSet) error {
+	calcSealData := PrepareSeal(header, round, sealType)
+	calcSealer, err := qbft.GetSignatureAddressNoHashing(calcSealData, seal)
+	if err != nil {
+		return err
+	}
+	if calcSealer.Cmp(sealer) != 0 {
+		return errInvalidSigner
+	}
+	if _, val := valSet.GetByAddress(sealer); val == nil {
```
jennwiederholen: checking msg.source is one of the valset is no need here since it is already done when message comes

jennwiederholen (commented):
Looks like the bug in VerifyBlockProposal and verifyHeader is still not solved!
One that skips verifying PrevCommittedSeal for BlockProposal and we had discussed verifying PrevSeals and verifyingCurrentSeals should be split up or something

consensus/qbft/core/core.go
egonspace: Okay, I'll apply it.

egonspace:
> Looks like the bug in VerifyBlockProposal and verifyHeader is still not solved! One that skips verifying PrevCommittedSeal for BlockProposal and we had discussed verifying PrevSeals and verifyingCurrentSeals should be split up or something

I didn't fixed it yet with this commit. I'll fix it with another commit.

consensus/qbft/backend/backend.go
```diff
@@ -265,12 +265,16 @@ func (sb *Backend) Verify(proposal qbft.Proposal) (time.Duration, error) {
...
 
 	header := block.Header()
-	snap, err := sb.snapshot(sb.chain, header.Number.Uint64()-1, header.ParentHash, nil)
-	if err != nil {
+	var snap, prevSnap *Snapshot
+	var err error
+
+	if snap, err = sb.snapshot(sb.chain, header.Number.Uint64()-1, header.ParentHash, nil); err != nil {
+		return 0, err
+	} else if prevSnap, err = sb.snapshot(sb.chain, header.Number.Uint64()-2, header.ParentHash, nil); err != nil {
```
egonspace, line 273: Shouldn't we call `snapshot` with the parentHash of "parent block"?

egonspace: However, since these codes will need to be modified once the epoch is implemented, let's just keep it in mind for future review.

egonspace (approved):

consensus/qbft/backend/engine.go
```diff
@@ -88,7 +88,7 @@ func (sb *Backend) verifyHeader(chain consensus.ChainHeaderReader, header *types
 	if snap, err = sb.snapshot(chain, header.Number.Uint64()-1, header.ParentHash, parents); err != nil {
 		return err
 	} else if header.Number.Uint64() < 2 {
-		return sb.Engine().VerifyHeader(chain, header, parents, snap.ValSet, snap.ValSet)
+		return sb.Engine().VerifyHeader(chain, header, parents, snap.ValSet, snap.ValSet, true)
```
jennwiederholen, line 91: shouldn't it be false here?

consensus/qbft/engine/engine.go
```diff
@@ -195,7 +195,7 @@ func (e *Engine) VerifyHeaders(chain consensus.ChainHeaderReader, headers []*typ
 			if errored {
 				err = consensus.ErrUnknownAncestor
 			} else {
-				err = e.verifyHeader(chain, header, headers[:i], validators, prevValidators)
+				err = e.verifyHeader(chain, header, headers[:i], validators, prevValidators, false)
```
jennwiederholen: why should we skip checking seals for verifying headers  here?

consensus/qbft/backend/engine.go
egonspace: Shouldn't it be called by only `VerifyBlockProposal` with false?

consensus/qbft/engine/engine.go
egonspace: It's my fault. But I found it's not used anywhere, so we can remove `engine:VerifyHeaders`. (`Backend.VerifyHeaders` does not call this function)

consensus/qbft/backend/engine.go
jennwiederholen: yes you're right. i got confused :)

PR #40, opened 2024-11-20
fix: wemix engine handler apis

egonspace:
- `qbft.Handler` apis should only be called after wbft engine has started.
- use unbuffered channel for new head subscription
- move `Node` struct used only for test
- fix `cmd` test data with hive tool
- fix `TestExtractToQBFTExtra` test
- fix test failures
  - `internal/ethapi`
  - `ethclient/simulated`
  - `consensus/qbft/validator`


jennwiederholen (dismissed):

consensus/qbft/engine/engine.go
```diff
@@ -316,9 +316,8 @@ func (e *Engine) verifyPreparedSeals(chain consensus.ChainHeaderReader, header *
 	prevPreparedSeal := extra.PrevPreparedSeal
 
 	if len(prevPreparedSeal) == 0 {
-		// prevPreparedSeal validation for monblanc block and block number 1
-		// is skipped because it's empty
-		if !(chain.Config().MontBlancBlock.Cmp(header.Number) == 0) || !(number == 1) {
+		// prevPreparedSeal validation for monblanc block or first block after genesis is skipped because it's empty
+		if chain.Config().MontBlancBlock.Cmp(header.Number) == 1 && header.Number.Uint64() != 1 {
```
jennwiederholen: chain.Config().MontBlancBlock.Cmp(header.Number) != 0 
would show the  code's  purpose better. 
minor one , however,  we can reuse  variable number for  header.Number.Uint64()

jennwiederholen (approved):

consensus/qbft/engine/engine.go
egonspace: It was my mistake. I fixed it.

PR #42, opened 2024-11-27
Update verify header for wbft

jed-wemade:
This PR makes wbft check header more striclty.
The checking logic is taken from clique implementation.


consensus/qbft/backend/engine_test.go
```diff
@@ -179,19 +180,22 @@ func copyConfig(config *qbft.Config) *qbft.Config {
 	return &cpy
 }
 
-func makeHeader(parent *types.Block, config *qbft.Config) *types.Header {
+// makeHeader create header executing no txs
+func makeHeader(chain *core.BlockChain, engine *Backend, parent *types.Block) *types.Header {
```
egonspace: Wouldn't it be better to get a narrower range of parameter(`engine.Config`)?

jed-wemade: Fixed.

egonspace (approved):

PR #43, opened 2024-11-27
feat: collect additional seals than  2f+1

### LINKED ISSUE #16: implement adding committed seal to next block

jennwiederholen:
related issue: https://github.com/wemixarchive/go-wemix-qbft/issues/16

The feature is to add additional  block n 's prepare and committed seal to block n+1 's prevPreparedSeal and prevCommittedSeal. The purpose is to motivate validators to run consensus diligently by giving them incentive even their message verification is not included in 2f+1 message because of the lazy network condition or something.

Additional seals than 2f+1  are collected in `core.extraSeals` and it is added to prevPrepared/CommittedSeal field when `backdend` `Prepare` new proposal (which is basically block)


consensus/qbft/backend/engine_test.go
```diff
@@ -141,12 +142,15 @@ func newBlockchainFromConfig(genesis *core.Genesis, nodeKeys []*ecdsa.PrivateKey
...
 	// find proposer key
-	for _, key := range nodeKeys {
+	for i, key := range nodeKeys {
 		addr := crypto.PubkeyToAddress(key.PublicKey)
 		if addr.String() == proposerAddr.String() {
 			backend.privateKey = key
 			backend.address = addr
 			backend.qbftEngine = qbftengine.NewEngine(backend.config, addr, backend.Sign)
+			// set backend's node to first index of nodes for convenient
+			nodes[0], nodes[i] = nodes[i], nodes[0]
```
egonspace, line 154: FYI: this swap code is safe in golang. (https://stackoverflow.com/questions/76399227/is-it-safe-to-swap-two-integers-by-a-b-b-a-in-golang)

consensus/qbft/core/backlog.go
```diff
@@ -99,10 +109,18 @@ func (c *Core) checkMessage(msgCode uint64, view *qbft.View) error {
 		// StatePrepared only accepts msgCommit and msgRoundChange
 		// other messages are invalid messages
 		if msgCode < qbftmessage.CommitCode {
+			if msgCode == qbftmessage.PrepareCode {
+				return errExtraSealMessage
+			}
 			return errInvalidMessage
 		}
```
egonspace:
```suggestion
		if msgCode == qbftmessage.PrepareCode {
			return errExtraSealMessage
		} else if msgCode < qbftmessage.CommitCode {
			return errInvalidMessage
		}
```

consensus/qbft/core/extraseal.go
```diff
@@ -0,0 +1,103 @@
...
+		if prepareMsg.Digest != c.current.Proposal().Hash() {
+			logger.Error("QBFT: invalid PREPARE message digest")
+			return errInvalidMessage
+		}
+
+		if err := verifySeal(block.Header(), uint32(prepareMsg.CommonPayload.Round.Uint64()), SealTypePrepare,
+			prepareMsg.PrepareSeal, prepareMsg.Source()); err != nil {
+			return errInvalidSeal
+		}
+	}
```
egonspace:
```suggestion
	if prepareMsg, ok := msg.(*qbftmessage.Prepare); ok {
		// Check digest
		if prepareMsg.Digest != c.current.Proposal().Hash() {
			logger.Error("QBFT: invalid PREPARE message digest")
			return errInvalidMessage
		}

		if err := verifySeal(block.Header(), uint32(prepareMsg.CommonPayload.Round.Uint64()), SealTypePrepare,
			prepareMsg.PrepareSeal, prepareMsg.Source()); err != nil {
			return errInvalidSeal
		}
	} else if commitMsg, ok := msg.(*qbftmessage.Commit); ok {
			// Check digest
			if commitMsg.Digest != c.current.Proposal().Hash() {
				logger.Error("QBFT: invalid COMMIT message digest")
				return errInvalidMessage
			}

			if err := verifySeal(block.Header(), uint32(commitMsg.CommonPayload.Round.Uint64()), SealTypeCommit,
				commitMsg.CommitSeal, commitMsg.Source()); err != nil {
				return errInvalidSeal
			}
	} else {
			return errInvalidExtraSealMessage
	}
```

jennwiederholen:
[d52e724](https://github.com/wemixarchive/go-wemix-qbft/pull/43/commits/d52e7247ee0ea78f3481474f5adf57df1f03e806)
Process ExtraSeals **right after block time delay is over, and before `qbft.Backend` sends Request to `qbft.core`.**
 
The option to process extraSeal when **making preprepare message** found out to be inappropriate because when `backend` sends request to `core` , it stores `proposalHash` and it is used when `core` reach `committed`. `proposalHash` is compared with `blockHash` that has reached committed . So if we change block hash ( by adding extraSeals) when preprepare message is created, it will always be different from `proposalHash`

consensus/qbft/core/core.go
```diff
@@ -95,7 +95,7 @@ type Core struct {
 
 	extraSeals   *prque.Prque[int64, qbftmessage.QBFTMessage]
 	extraSealsMu *sync.Mutex
-	priorRound   *big.Int // latest round that is committed
+	priorState   *priorState
```
egonspace: If we don't want to check whether this field is nil and it doesn't need to be nil, I think there's no need to declare it as a pointer type.

consensus/qbft/core/priorstate.go
```diff
@@ -0,0 +1,36 @@
...
+// priorState stores prior consensus state
+// for collecting extra seals during StateAcceptRequest
+type priorState struct {
+	mu       *sync.RWMutex
+	round    *big.Int
+	proposal qbft.Proposal
+}
+
+func (c *Core) updatePriorState(priorRound *big.Int, priorProposal qbft.Proposal) {
+	c.priorState.round = priorRound
```
egonspace, line 21: I think these two settings need to be protected by a mutex lock.

jennwiederholen: right, i missed it😅

consensus/qbft/core/core.go
jennwiederholen: yes the priorState cannot be nil in any case. i applied your comment.

jennwiederholen:
This version has several problems to be fixed
: 
1. the consensus will not work due to "bad block" error because the block rewardee is changed after block is prepared and header.Root is set due to extraSeal logic
2. "Block found but no relative pending task" error occurs because SealHash() is different from the block that node preapred and the block passed by result channel

above problem will be fixed after https://github.com/wemixarchive/go-wemix-qbft/pull/44 is merged , by moving the execution point of processing extraseal to the time when the block is prepared

consensus/qbft/backend/engine_test.go
```diff
@@ -879,8 +883,119 @@ func TestMakingBlock(t *testing.T) {
...
+		defer wg.Done()
+		for {
+			<-ticker.C
+			consensusState := engine.core.GetState().String()
+			if executed[consensusState] {
+				continue
+			}
+
+			switch consensusState {
+			case "Preprepared":
```
jed-wemade: Using defined constant as itself looks good like `qbftcore.StatePreprepared`.

consensus/qbft/backend/engine.go
```diff
@@ -250,6 +251,13 @@ func (sb *Backend) Seal(chain consensus.ChainHeaderReader, block *types.Block, r
 			return
 		}
 
+		// process extraSeals add apply new header to block
+		headerWithExtraSeals, err := sb.processExtraSeals(block.Header())
```
egonspace: I think `processExtraSeals` does not need to be in a go routine. How about to put it just before `Seal()`?

consensus/qbft/backend/engine_test.go
jennwiederholen: agree with it :) i applied it

consensus/qbft/backend/engine.go
jennwiederholen: it's there because processingExtraSeal had to be done after block period delay . however the block delay timing is changed, and also processing timing  will be moved to the point when block is prepared anyway.

consensus/qbft/core/extraseal.go
```diff
@@ -0,0 +1,110 @@
...
+			return errInvalidSeal
+		}
+	} else {
+		return errInvalidExtraSealMessage
+	}
+
+	c.extraSealsMu.Lock()
+	defer c.extraSealsMu.Unlock()
+	view := msg.View()
+	c.extraSeals.Push(msg, toPriority(&view))
```
egonspace: It seems more efficient to handle duplicate seals when adding seals rather than in `ProcessExtraSeal`. During BFT consensus, many duplicate seals are likely to be received, and there’s no need for many identical seals to accumulate at the time of adding.

jennwiederholen: given comment is applied to new commit :) 
ExtaSeal is saved as map, and only single extraSeal is saved per message signer. so duplicated message will be discarded when adding extraSeals.

consensus/qbft/core/core.go
```diff
@@ -90,6 +93,10 @@ type Core struct {
 	backlogs   map[common.Address]*prque.Prque[int64, qbftmessage.QBFTMessage]
 	backlogsMu *sync.Mutex
 
+	extraSeals   map[common.Address]map[SealType]qbftmessage.QBFTMessage
```
egonspace: A map of maps seems a bit complex so how about to define two fields separately like `extraPrepareSeals` and `extraCommitSeals`?

egonspace: Otherwise, how about this
```
extraSeals map[SealType]map[common.Address]qbftmessage.QBFTMessage
```
and, we never delete `extraSeals[SealTypePrepare]` and `extraSeals[SealTypeCommit]`?

jennwiederholen: Pros of Inner map struct is that it reduces the complexity of code but in second thought, managing two separate map will be memory efficient as we always have only two kinds  of SealType. So i'll separte the inner map into two seprate maps :)

jennwiederholen: fixed in 
[65dbcab](https://github.com/wemixarchive/go-wemix-qbft/pull/43/commits/65dbcab199cc9d30387392571d441341ed127aed)

consensus/qbft/core/core.go
```diff
@@ -61,9 +61,13 @@ func New(backend Backend, config *qbft.Config) *Core {
 		backend:            backend,
 		backlogs:           make(map[common.Address]*prque.Prque[int64, qbftmessage.QBFTMessage]),
 		backlogsMu:         new(sync.Mutex),
+		prepareExtraSeals:  make(map[common.Address]*qbftmessage.Prepare),
+		commitextraSeals:   make(map[common.Address]*qbftmessage.Commit),
```
egonspace: typo: `commitExtraSeals`

consensus/qbft/backend/engine.go
```diff
@@ -282,6 +285,16 @@ func (sb *Backend) Seal(chain consensus.ChainHeaderReader, block *types.Block, r
 	return nil
 }
 
+func (sb *Backend) processExtraSeals() (map[common.Hash][]byte, map[common.Hash][]byte) {
+	if sb.core == nil {
+		return nil, nil
+	} else {
+		lastProposal, _ := sb.LastProposal()
```
egonspace: minor: you can use simply `sb.currentBlock()` instead of `sb.LastProposal()`

jennwiederholen: that's better!

consensus/qbft/core/extraseal.go
```diff
@@ -0,0 +1,120 @@
...
+		sealType SealType
+	)
+
+	if c.state == StateAcceptRequest {
+		block, ok = c.priorState.Proposal().(*types.Block)
+	} else {
+		block, ok = c.current.Proposal().(*types.Block)
+	}
+	if !ok {
+		return errInvalidMessage
```
egonspace: This error doesn't seem to fit the situation very well. I think `priorState.Proposal()` can be nil in special case(ex; in making first block of a chain), so if `ok` is false, then we can just ignore the `msg` without error.

consensus/qbft/core/handler.go
```diff
@@ -199,6 +199,11 @@ func (c *Core) handleDecodedMessage(m qbftmessage.QBFTMessage) error {
 		if err == errFutureMessage {
 			c.addToBacklog(m)
 		}
+		if err == errExtraSealMessage {
```
egonspace, line 202: I think we can `addToExtraSeal(m)` at the `processBacklog()` as well.

consensus/qbft/core/extraseal.go
jennwiederholen: fixed in [e6c1785](https://github.com/wemixarchive/go-wemix-qbft/pull/43/commits/e6c1785a86e6a9b6d38d318b9c6189dacd2228f0)

consensus/qbft/core/handler.go
jennwiederholen: fixed in [8190450](https://github.com/wemixarchive/go-wemix-qbft/pull/43/commits/8190450fcdbd85109d8d05609dd5d97c50f65c5f)

egonspace (approved):

PR #44, opened 2024-12-03
fix: delayed commit new work for qbft

egonspace:
- as is: Worker commits a new work just after new head has committed. So the qbft engine waits for a block period in `Seal()`.
- to be: Worker waits a block period before committing a new work so as it could `Prepare` a block including extra seals as many as possible
- other engine and simulated engine is not effected.
- etc: revoke wemix logic from `CalcGasLimit`

### DIFF
--- a/consensus/qbft/backend/engine.go
+++ b/consensus/qbft/backend/engine.go
@@ -161,6 +161,13 @@ func (sb *Backend) VerifySeal(chain consensus.ChainHeaderReader, header *types.H
 	return sb.Engine().VerifySeal(chain, header, snap.ValSet)
 }
 
+func (sb *Backend) TimeToNextBlock() uint64 {
+	latestBlock := sb.currentBlock()
+	next := new(big.Int).Set(latestBlock.Number())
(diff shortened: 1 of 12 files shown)


consensus/qbft/backend/engine.go
```diff
@@ -161,6 +161,19 @@ func (sb *Backend) VerifySeal(chain consensus.ChainHeaderReader, header *types.H
 	return sb.Engine().VerifySeal(chain, header, snap.ValSet)
 }
 
+func (sb *Backend) TimeToNextBlock() uint64 {
+	if sb.currentBlock == nil {
+		return 0 // if it has no current block function then it returns zero time
+	}
+	latestBlock := sb.currentBlock()
```
paul-ahn-wm, line 168: isn't it same with codes above?
I think lines 169 - 171 are needless

egonspace: We need to check whether `currentBlock` function is nil, as well as whether `currentBlock()` function return value is nil.

jennwiederholen (approved):

PR #54, opened 2024-12-19
feat: modify newWorkLoop for WBFT

kai-yu-wm:
Restore the default newWorkerLoop and create a newWorkerLoop specifically for WBFT.

### DIFF
--- a/miner/worker.go
+++ b/miner/worker.go
@@ -447,17 +447,119 @@ func recalcRecommit(minRecommit, prev time.Duration, target float64, inc bool) t
 
 // newWorkLoop is a standalone goroutine to submit new sealing work upon received events.
 func (w *worker) newWorkLoop(recommit time.Duration) {
+	if w.chainConfig.QBFT == nil {
+		w.newWorkLoopOrigin(recommit)
+	} else {
+		w.newWorkLoopWBFT()
+	}
+}
+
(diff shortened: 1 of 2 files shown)


miner/worker.go
```diff
@@ -579,6 +552,97 @@ func (w *worker) newWorkLoop(recommit time.Duration) {
...
+		if w.config.SimulatedEnabled {
+			delayTimer.Reset(0)
+		} else {
+			// There is some engine in which worker needs to wait until the next block time to commit new work(ex: qbft).
+			// In the previous QBFT engine, worker should wait for the block period when sealing.
+			// However, to create a finalized block(including extra seals) during the prepare phase, it is necessary
+			// to wait for the block period when processing new work. The reason for waiting during the prepare phase is
+			// to gather more extra seals.
+			// If another `tryCommit` call happens before delayTimer tick occurs, prior timer is discarded.
+			delayTimer.Reset(time.Until(time.Unix(int64(w.engine.TimeForNextWork()), 0)))
```
egonspace, line 592: For more safe processing, when we reset the timer we should flush the channel if the timer already has been up to the time so as we could defend duplicate `committing`.
```
if !delayTimer.Stop() {
    <-delayTimer.C
}
delayTimer.Reset(...)
```

colinkim:
Can we leave the part that allows recommit to be used externally (CLI/web3, etc.)?

PR #56, opened 2024-12-24
feat: implement epoch block

Noah-jhkim:
Introduced the concept of an Epoch, representing a specific cycle of block generation.
- Defined the EpochBlock as the block that records the ValidatorList for the Nth Epoch.
- Specified that the EpochBlock for the (N)th Epoch is the last block of the (N-1)th Epoch.

**Changes**
- Added functions to support Epoch and EpochBlock concepts.
- Added handling process for EpochBlock in the finalize process.
- Added a validation process for EpochBlock validators within the insertChain process.
- Removed the snapshot function and related functions.
- Enhanced code readability (fn `verifyHeader`)
- Remove `EmptyBlockPeriod` config and related logic
- Remove `Vote` from Extra
- Remove sort from `ValidatorSet`


consensus/qbft/config.go
```diff
@@ -202,6 +204,122 @@ func (c *Config) getTransitionValue(num *big.Int, callback func(transition param
...
+// - It can be nil if the nearest hard fork block number corresponding to the given block number does not exist.
+func (c *Config) getNearestForkBlock(blockNumber *big.Int) *big.Int {
+	// TODO: (noah) transition의 Block 순서를 보장하지 못하는 경우를 가정하여 만듬
+	var nearestForkBlock *big.Int = nil
+
+	for _, transition := range c.Transitions {
+		if transition.Block.Cmp(blockNumber) == 0 {
+			return transition.Block
+		}
+		if transition.Block.Cmp(blockNumber) < -1 && transition.Block.Cmp(nearestForkBlock) > 0 {
```
jennwiederholen: I believe this is what you intended 😃

```suggestion
		if transition.Block.Cmp(blockNumber) < 0 && transition.Block.Cmp(nearestForkBlock) > 0 {
```

Noah-jhkim: good!!

core/epoch_validator.go
```diff
@@ -0,0 +1,45 @@
...
+		for i := 0; i < len(addrs); i++ {
+			for j := i + 1; j < len(addrs); j++ {
+				if bytes.Compare(addrs[i][:], addrs[j][:]) > 0 {
+					addrs[i], addrs[j] = addrs[j], addrs[i]
+				}
+			}
+		}
+	}
+	sort(validatorFromState)
+	sort(validators)
```
jed-wemade: It looks like using `slices.SortFunc` is enough to do.
Also, validators in extra field should be already sorted so `validators` should not be sorted in this scope to verify the policy.

```go
slices.SortFunc(validatorsFromState, common.Address.Cmp)
// do not sort for `validators`
```

core/epoch_validator.go
```diff
@@ -0,0 +1,45 @@
...
+
+	// Checks if two arrays have the same elements in the same order
+	{
+		// Check if the lengths are different
```
jed-wemade: Lengths can be compared before sorting.

consensus/qbft/engine/engine.go
```diff
@@ -854,3 +850,18 @@ func mergeSeals(seals [][]byte, extraSeals map[common.Hash][]byte) [][]byte {
 	}
 	return mergedSeals
 }
+
+func writeValidatorsToEpoch(header *types.Header, state govwbft.StateReader) error {
+	return ApplyHeaderQBFTExtra(header, WriteValidators(ethcore.GetValidatorsFromState(state)))
+}
+
+// DefaultEpochHandler is a handler that performs default actions when the block is an EpochBlock,
```
jed-wemade: `DefaultEpochHandler` -> `verifyEpoch`

core/blockchain.go
```diff
@@ -1773,6 +1773,25 @@ func (bc *BlockChain) insertChain(chain types.Blocks, setHead bool) (int, error)
 		}
 		ptime := time.Since(pstart)
 
+		// Checking if the headers contains at least one MontBlanc block.
+		if bc.chainConfig.IsMontBlanc(headers[len(headers)-1].Number) || bc.chainConfig.MontBlancBlock == nil {
```
jed-wemade: This makes `VerifyValidators()` run twice if the chain is WBFT chain or the block is MontBlanc block on WEMIX chain. (once from `Finalize()` in WBFT engine)
Is this intended behavior?

consensus/qbft/backend/engine.go
```diff
@@ -82,37 +80,49 @@ func (sb *Backend) VerifyHeader(chain consensus.ChainHeaderReader, header *types
...
-		return sb.Engine().VerifyHeader(chain, header, parents, snap.ValSet, snap.ValSet, true)
-	} else if len(parents) < 2 {
-		var parent *types.Header
-		if len(parents) == 1 {
-			parent = parents[0]
+	// Retrieve the ValidatorSet of block
+	parentNum := new(big.Int).Set(header.Number)
+	parentNum = parentNum.Sub(parentNum, common.Big1)
+	if valSet, err = sb.GetValidators(chain, parentNum, header.ParentHash); err != nil {
+		return consensus.ErrUnknownAncestor
```
jed-wemade: `consensus.ErrUnknownAncestor` -> `err`

core/blockchain.go
egonspace: You're right. It seems like the `VerifyValidators` here is unnecessary.

core/epoch_validator.go
egonspace: applied

consensus/qbft/engine/engine.go
egonspace: applied

consensus/qbft/engine/engine.go
```diff
@@ -623,6 +620,38 @@ func (e *Engine) ExtractGenesisValidators(header *types.Header) ([]common.Addres
...
+			break
+		}
+		// EPOCH RULE: all epoch transition blocks are an epoch block
+		firstNewEpoch.Set(transition.Block)
+		epochLength.SetUint64(transition.EpochLength)
+	}
+	num := number.Uint64()
+	first := firstNewEpoch.Uint64()
+	epochLen := epochLength.Uint64()
+	return (num-first)%epochLen == 0, new(big.Int).SetUint64((num-first)/epochLen*epochLen + first), nil
```
jed-wemade:
```suggestion
	rem := new(big.Int).Sub(number, firstNewEpoch)
	rem.Rem(rem, epochLength)
	return rem.Sign() == 0, new(big.Int).Sub(number, rem), nil
```

consensus/qbft/engine/engine.go
```diff
@@ -623,6 +620,38 @@ func (e *Engine) ExtractGenesisValidators(header *types.Header) ([]common.Addres
...
+	epochLength := new(big.Int).SetUint64(e.cfg.Epoch)
+	firstNewEpoch := new(big.Int).SetUint64(0)
+	if config.MontBlancBlock != nil {
+		firstNewEpoch.Set(config.MontBlancBlock)
+	}
+	for _, transition := range e.cfg.Transitions {
+		if config.MontBlancBlock != nil {
+			if transition.Block.Cmp(config.MontBlancBlock) < 0 {
+				// skip transitions before MontBlancBlock; it should be ignored
+				continue
```
jed-wemade: `GetConfig()` for QBFT is always applied regardless of hard fork. Is `EpochLength` exceptional?

egonspace: The code was added to handle the case where there’s a QBFT config transition before the MontBlanc hard fork, as it’s an unusual setup, but I’ll remove it for consistency.

consensus/qbft/validator/default_test.go
```diff
@@ -59,20 +67,11 @@ func testNewValidatorSet(t *testing.T) {
 	}
 
 	// Create ValidatorSet
-	valSet := NewSet(ExtractValidators(b), qbft.NewRoundRobinProposerPolicy())
+	valSet := NewSet(extractValidators(b), qbft.NewRoundRobinProposerPolicy())
```
jed-wemade: Use address list directly.
```go
var validators []common.Address
for i := 0; i < ValCnt; i++ {
	key, _ := crypto.GenerateKey()
	validators = append(validators, addr)
}
valSet := NewSet(validators, qbft.NewRoundRobinProposerPolicy())
```

consensus/qbft/backend/engine.go
```diff
@@ -21,30 +21,24 @@
...
 	"github.com/ethereum/go-ethereum/rpc"
 )
 
 const (
-	checkpointInterval = 1024 // Number of blocks after which to save the vote snapshot to the database
-	inmemorySnapshots  = 128  // Number of recent vote snapshots to keep in memory
-	inmemoryPeers      = 40
-	inmemoryMessages   = 1024
+	inmemoryPeers    = 40
+	inmemoryMessages = 1024
```
jed-wemade, line 41: Are these parameters fixed?
Large enough for `inmemoryMessages` but not for `inmemoryPeers` perhaps.

consensus/qbft/backend/backend.go
```diff
@@ -319,20 +304,16 @@ func (sb *Backend) GetProposer(number uint64) common.Address {
...
+func (sb *Backend) Validators(proposal qbft.Proposal) qbft.ValidatorSet {
+	valSet, err := sb.Engine().GetValidators(sb.chain, new(big.Int).Add(proposal.Number(), common.Big1), proposal.Hash(), nil)
 	if err != nil {
 		return validator.NewSet(nil, sb.config.ProposerPolicy)
 	}
-	return snap.ValSet
+	return valSet
+}
+
+func (sb *Backend) CurrentProposer() common.Address {
```
jed-wemade: `CurrentProposer()` is an ambiguous name because it is expected as if it is identical to `Backend.GetProposer(-1)`.
`ProposerFromValSet()` or something would be good.

consensus/qbft/engine/engine.go
```diff
@@ -623,12 +609,102 @@ func (e *Engine) ExtractGenesisValidators(header *types.Header) ([]common.Addres
...
+	if len(parents) > 0 {
+		block = parents[len(parents)-1]
+		parents = parents[:len(parents)-1]
+	} else {
+		block = chain.GetHeader(parentHash, parentNumber.Uint64())
+	}
+	if block == nil {
+		return nil, consensus.ErrUnknownAncestor
+	}
+	for !isEpoch && block.Number.Cmp(latestEpoch) > 0 {
```
jed-wemade, line 677: For epoch block, validator set in `valSetCache` is not used and updated to new validator set.
Due to cache working in LRU manner, the update makes cache fresh but I don't know it is intended for this context.

consensus/qbft/backend/engine.go
egonspace: It’s not subject to a hard fork, so each node can increase it as needed.

jed-wemade (approved):

PR #58, opened 2024-12-30
doc: introduce WBFT

egonspace:


README.md
```diff
@@ -1,16 +1,220 @@
...
+
+- Adoption of DPoS: Allows anyone to participate as a validator through staking
+- Validator selection: Chosen via VRF based on staking amount and validation diligence
+- Reward system and diligence metrics
+- Concept of epoch: Defines a unit where the validator set changes
+- Inclusion of consensus proof in the agreement process
+- Improved miner worker operations
+- Enhanced block header verification
+
+### Terminology
```
jed-wemade, line 13: Recurring explanation or re-defining terminology without using "same as IBFT" or "unlike IBFT".

README.md
```diff
@@ -1,16 +1,220 @@
...
+  - The maximum value of `d` is 2
+  - The minimum value of `d` is 0 (no proposals or seals)
+- ex) if `e=200, v=20, w=10` then `d = p / 400 + s / 400`
+
+When a proposer writes the diligence in a epoch block, it uses accumulated diligence for each staker.
+
+Cumulative diligence `D(n) = D(n-1) * 0.9 + d(n) * 0.1`
+- `D(n)` is the cumulative diligence until the n-th epoch
+- `d(n)` is the diligence of the n-th epoch
+- `D(n-1)` is the cumulative diligence until the (n-1)th epoch. If it becomes a staker at first, its `D(n-1) = 1.9` (default)
```
jed-wemade: > `D(n-1) = 1.9` (default)

`D(n-1) = 1.9` (default. 95% of the maximum value of diligence)

README.md
```diff
@@ -1,16 +1,220 @@
...
+to provide rewards for the validation role as well. However, direct coin rewards based on validation actions are avoided 
+due to potential misuse (e.g., block proposers deliberately omitting a specific validator's consensus proof). 
+Instead, validation rewards are indirectly tied to validator selection. Validators who diligently perform their 
+validation duties have a better chance of being selected and can earn greater rewards in this structure.
+
+Diligence is calculated at the end of each epoch:
+- Let `e` be the number of blocks in the epoch
+- Let `v` be the number of validators
+- Let `w` be the number of blocks proposed by a validator during the epoch
+  - round change may be occured so `w` is not always equal to `e/v`
```
jed-wemade: Despite of no round change, `e` may not be equal to `v*w` because `v` is independent to `e`.

README.md
```diff
@@ -1,16 +1,220 @@
...
+- `d(n)` is the diligence of the n-th epoch
+- `D(n-1)` is the cumulative diligence until the (n-1)th epoch. If it becomes a staker at first, its `D(n-1) = 1.9` (default)
+  - If the default value is too low, the probability of being selected as a validator when first becoming a staker will be low. Conversely, if it is too high, it may be advantageous to become a new staker again after even minor mistakes. Therefore, an appropriate value is necessary.
+
+### Concept of Epoch
+The WBFT configuration allows defining the size of an epoch. An epoch represents the period (in terms of block count) during which a predetermined validator set remains active. The last block of an epoch is referred to as an epoch block. The genesis block is considered an epoch block; hence, the first epoch starts from block 1. When the proposer suggests a block that is an epoch block, the following steps are performed:
+
+- Reflect the diligence shown by the current validator set during this epoch in their cumulative diligence and record it in the extra field of the block header (if an staker was not part of the validator set during this epoch, its cumulative diligence is not updated)
```
jed-wemade: Cumulative diligence is updated for all stakers regardless of whether staker has validator role.

> (if an staker was not part of the validator set during this epoch, its cumulative diligence is not updated)

Moving to diligence calculation.

egonspace: "Cumulative diligence updating" issue needs to be explained in "Epoch Concept" chapter.

README.md
```diff
@@ -1,16 +1,223 @@
...
+
+- `Previous Prepare Seal`: The prepare seal for the previous block. It is included in the block hash and consensus.
+- `Previous Commit Seal`: The commit seal for the previous block. It is included in the block hash and consensus.
+- `Prepare Seal`: The prepare seal for the current block. It is not included in the block hash or consensus.
+- `Commit Seal`: The commit seal for the current block. It is not included in the block hash or consensus.
+
+When a validator becomes the proposer, they create the current block by combining the prepare and commit seals stored in the previous block with any additional prepare and commit messages received. These are recorded in the current block as the `Previous Prepare Seal` and `Previous Commit Seal`.
+
+The prepare and commit seals stored in the previous block are messages collected from peers just before the block is finalized. However, before the next block is created, the block period allows additional prepare and commit messages to be received. WBFT improves upon IBFT/QBFT by aiming to collect as many of these messages as possible, rather than stopping at just one-third.
```
jed-wemade: > rather than stopping at just one-third

"one-third" -> "two-third"

README.md
```diff
@@ -1,16 +1,223 @@
...
+	CommittedSeal               [][]byte
+	AggregatedPrevPreparedSeal  []byte
+	AggregatedPrevCommittedSeal []byte
+	Epoch                       EpochInfo
+}
+```
+
+### WEMIX 3.5
+
+WEMIX 3.5 defines a hard fork from the existing WEMIX PoA consensus to the WBFT consensus and adopts an intermediate consensus method with some features disabled for a safe transition to WBFT.
```
jed-wemade: For clarity, "PoA" -> "SPoA"

README.md
```diff
@@ -1,16 +1,223 @@
...
+
+#### MontBlanc hard fork
+
+WBFT is designed to record the first epoch information in the genesis block. However, since the WEMIX chain needs to transition from an already existing chain to a WBFT chain, exception rules for this are defined in the mont blanc hard fork. The following protocol applies to this hard fork:
+- WPoA validator nodes stop block creation when it is time to create the mont blanc block (i.e., they create blocks up to just before the mont blanc hard fork).
+- WBFT validator nodes receive block propagation normally up to the block before mont blanc.
+- WBFT validator nodes recognize the mont blanc hard fork and can obtain the validator set from the WBFT config when it is their turn to create this block.
+- WBFT validator nodes can create blocks and proceed with consensus once they can obtain the validator set.
+- The mont blanc block is a special block. Validators that receive this block perform additional tasks to deploy and initialize the NCP contract and GovStaking contract.
+- The mont blanc block is the first epoch block of WBFT. Therefore, the minimum validator set is recorded.
```
jed-wemade: > Therefore, the minimum validator set is recorded.

"minimum validator set" is not mentioned anywhere.
Is this the minimum set to run WBFT? (i.e., 4 validators to allow 1 byzantine node)

README.md
```diff
@@ -1,16 +1,223 @@
...
+- The first epoch starts from the block after the mont blanc block, during which stakers start staking from zero.
+- If the number of stakers is equal to or greater than the minimum stakers during the first epoch, these stakers become validators from the next epoch.
+- If the number of stakers is less than the minimum stakers during the first epoch, the initial validator set is maintained.
+
+#### NCP
+
+Although WBFT is designed and implemented to be used as a public chain, the Wemix chain executes WBFT consensus in the existing PoA method for a safer transition to a public chain. To this end, a group called NCP is defined, and only these NCPs can become validators. NCPs are selected from the existing Wemix 3.0 NCPs and are defined by contract. They have the obligation to maintain WEMIX 3.5 safely. The addition/removal of NCPs is decided by voting among NCPs, so it can proceed without a separate hard fork. Anyone can know the NCP list by querying the NCP contract. The NCP system is a temporary feature used only in WEMIX 3.5 and will not be used in WEMIX 4.0.
```
jed-wemade: > the Wemix chain executes WBFT consensus in the existing PoA method for a safer transition to a public chain.

"PoA method" is weird representation for me.

"the Wemix chain continues selecting validators from NCPs for a safer transition to a public chain as an intermediate step."

> They have the obligation to maintain WEMIX 3.5 safely.

"They have the obligation to run (mining) nodes for maintaining WEMIX 3.5 safely."

README.md
```diff
@@ -1,16 +1,223 @@
...
 
-Golang execution layer implementation of the Ethereum protocol.
+During the WEMIX 3.5 period, the validator set selection rules are as follows:
```
jed-wemade: > During the WEMIX 3.5 period

"period -> phase" OR "In WEMIX 3.5"

README.md
```diff
@@ -1,16 +1,223 @@
...
-[![API Reference](
-https://pkg.go.dev/badge/github.com/ethereum/go-ethereum
-)](https://pkg.go.dev/github.com/ethereum/go-ethereum?tab=doc)
-[![Go Report Card](https://goreportcard.com/badge/github.com/ethereum/go-ethereum)](https://goreportcard.com/report/github.com/ethereum/go-ethereum)
-[![Travis](https://app.travis-ci.com/ethereum/go-ethereum.svg?branch=master)](https://app.travis-ci.com/github/ethereum/go-ethereum)
-[![Discord](https://img.shields.io/badge/discord-join%20chat-blue.svg)](https://discord.gg/nthXNEv)
+The process of obtaining the validator set at any block height is as follows (replacing the snapshot function in the existing IBFT):
+- If the current block is an epoch block, obtain the validator set from the current block.
+- If the current block is not an epoch block, go back from the height - 1 block to find the most recent epoch block and obtain the validator set.
+- If you go back in time and encounter the mont blanc hard fork block or the genesis block, it is a failure if you do not meet them.
```
jed-wemade: > go back from the height - 1 block

"traverse blocks backward"

> it is a failure if you do not meet them.

"obtain the validator set from that block."

README.md
```diff
@@ -1,16 +1,223 @@
-## Go Ethereum
+## WBFT Protocol Specification (WEMIX 4.0)
+
+WBFT(WEMIX Byzantine Fault Tolerant) is a consensus algorithm that emphasizes decentralization, adapting Istanbul BFT(https://github.com/ethereum/EIPs/issues/650) and QBFT(https://github.com/Consensys/qbft-formal-spec-and-verification) for use in public blockchains. The following improvements have been implemented:
+
+- Adoption of DPoS: Allows anyone to participate as a validator through staking.
```
jed-wemade: For clarity, "DPoS" -> "DPoS(Delegated Proof of Stake)"

README.md
egonspace: In QBFT, a validator may include only F + 1 seals.
```
if validSeal <= validators.F() {
	return qbftcommon.ErrInvalidCommittedSeals
}
```

jed-wemade: Validation needs F + 1 seals, but collecting process needs two-third to make quorum.

README.md
egonspace: It was a typo. I'll fix it.

README.md
egonspace: The implementation was designed to collect more than 2/3 of the seals, but since the **consensus protocol** _allows_ committing as long as more than 1/3 of the seals are collected, it was originally described as 1/3. However, upon reviewing the sentence, it seems more appropriate to use 2/3 in this context, so it has been corrected accordingly.

README.md
```diff
@@ -1,16 +1,224 @@
...
+
+Diligence is calculated at the end of each epoch:
+- Let `e` be the number of blocks in the epoch.
+- Let `v` be the number of validators.
+- Let `w` be the number of blocks proposed by a validator during the epoch.
+- Let `p` be the total seals (prepare and commit) included in the proposed blocks by a validator.
+  - p can be equal to `2*v*w` at most.
+- Let `s` be the seals submitted by the validator during the epoch.
+  - s can be equal to `2*e` at most.
+- Diligence `d = p / (2*v*w) + s / (2*e)`.
```
jennwiederholen, line 95: As far as i understand, Diligence composed of 2 parts : 1. You diligently collected seals as proposer ( 50%) 2. You diligently submitted  as validator (50%) 
I see no problem if the validator is selected as proposer at least once in one epoch but if none, the diligence rapidly reduced to 50% at max.  And there's minor concern that it may be disadvantageous to whom that has never been selected as proposer, and it may be worse when the diligence is accumulated. 
Maybe we can apply adjusting variant α , for example  like .. : d=   α * ( p / (2*v*w) ) + (1-α) * ( s / (2*e) ) , (0 < α < 1)

egonspace: Any validator cannot go through an entire epoch without being selected as a proposer at least once. In other words, `w` cannot be zero.

jennwiederholen: In normal case, yes, every validator will be selected at lease once. but i think simple round-robin cannot consider the case for round-change due to network delay or something and w can be 0. If it's not, or if you think it's too minor to consider, let me know please :)

jennwiederholen: Definition of w is fixed and w cannot be 0. The comment is resolved.

jed-wemade: If `w = 0`, then `d = s / (2*e)` because the staker has no proposer role for that epoch but has validator role.
Furthermore, to make maximum value of `d` as 2, `d = s / e` would be good.

jed-wemade: > Any validator cannot go through an entire epoch without being selected as a proposer at least once.

If validator set becomes larger than epoch, `w` can be zero.
So epoch size should be larger enough (epoch size >= targetvalidators) to ensure every validator can propose blocks at least once.

README.md
```diff
@@ -1,16 +1,226 @@
...
+- First staking must exceed the minimum staking threshold.
+- No restrictions on amounts for additional staking.
+- Validators engaging in malicious behavior may be slashed, reducing their staking amount.
+- After unstaking, the remaining staking amount must either exceed the minimum threshold or be zero.
+- If the staking amount reaches zero, the staker is removed from staker set.
+- Unstaking releases funds to the staker’s address after the unbonding period.
+- Funds in the unbonding state can still be slashed.
+- Delegation is open to anyone, increasing the staking power of the recipient staker.
+- `Staking power`: Defined as staking amount + delegated amount.
+- Staking power determines both validation rewards and the probability of validator selection.
```
jennwiederholen:
```suggestion
- Staking
  - Initial staking amounts are valid at least the minimum staking threshold, and subsequent staking allows any amounts.
  - The valid stakers are on staker set but can be removed due to unstaking or slashing.
  - The changed staked amounts must be kept either at least the minimum staking threshold or 0.
  - Unstaked funds are released to the _staker operator address_ after the unbonding period (1 weeks).
  - Staking power for each staker is proportional to their own staked amounts, which determines both validation rewards and the probability of validator selection in the consensus process.
- Slashing
  - Stakers considered as malicious may be slashed, paying their own staked amounts completely to WEMIX ecosystem.
  - Delegated amounts are refunded to each delegator.
  - Stakers on the unbonding state can also be slashed.
- Delegation
  - Anyone can delegate funds to the staker with no amount limit.
  - Delegated amounts are credited to the recipient staker's staked amounts.
  - Undelegated funds are released to the _delegator address_ after the unbonding period (72 hours).
```

simple suggestion for organizing the content :)

README.md
```diff
@@ -1,16 +1,226 @@
...
+### Validator Selection
+In WBFT, the proposer of the last block in an epoch (referred to as the epoch block) selects the validator set for the next epoch and records it in the block. Validator selection rules:
+- The first validator set is defined in genesis.json.
+- Validators in the genesis block initially have a staking amount of zero.
+- First epoch is in a `stabilization stage`.
+- An epoch in a `stabilization stage` has the validator set which is same to the previous epoch.
+- If current epoch is in a `stabilization stage` and the number of stakers is below the minimum stakers in last block of an epoch, next epoch will be in a `stabilization stage`.
+- If the number of stakers is at least minimum stakers, validators are selected from stakers.
+- If the number of stakers equals or exceeds minimum stakers but is less than or equal to target validators, all stakers become validators.
+- If the number of stakers exceeds target validators, validators are selected using VRF, considering staking power and diligence.
```
jennwiederholen:
```suggestion
- Stabilization stage
  - The stabilization stage is the period from the first epoch until just before the first epoch when the number of stakers reaches the minimum required.
  - If current epoch is in a stabilization stage and the number of stakers is below the minimum stakers in last block of an epoch, next epoch will be in a stabilization stage.
  - An epoch in a stabilization stage has the validator set which is same to the previous epoch.
  - The very first validator set is defined in genesis.json, and validators in the genesis block initially have a staking amount of zero. 
- After stabilization stage , validator selection follows below rules 
  -  `minimum stakers <= number of stakers <= target validators` : every staker becomes validator 
  -  `number of stakers > target validators` : validators are selected using VRF, considering staking power and diligence.
  - `number of stakers < minimum stakers` : all remaining stakers becomes validator, which should not occur in a public network after stabilization stage for the sake of network security.
```

jed-wemade (approved):

PR #28, opened 2024-10-21
feat: implement wbft-simulated

kai-yu-wm:
Implement SimulatedBackend for qbft engine.

PR #60, opened 2025-01-13
feat: implement diligence, reward and beneficiaries

jed-wemade:
Changeset:
- Implement epoch info field in QBFT extra.
- Implement diligence score.
- Fix reward system based on staking amounts for validator set which is determined for each epoch.
- Add `blockRewardBeneficiaries` to reward Maintanence and Ecosystem in QBFT configs.
- Add `minStakers`, `targetValidators` for transition of validatior selection in QBFT configs.
- Remove `Votes` and `Validators` fields in QBFT extra.
- Remove unused QBFT configs.


consensus/qbft/core/request.go
```diff
@@ -50,55 +46,11 @@ func (c *Core) handleRequest(request *Request) error {
...
-			// Send PRE-PREPARE message to other validators
-			c.sendPreprepareMsg(request)
-		} else { // emptyBlockPeriod is set
-			c.newRoundMutex.Lock()
-			defer c.newRoundMutex.Unlock()
-
-			if c.newRoundTimer != nil {
-				c.newRoundTimer.Stop()
-				c.newRoundTimer = nil
-			}
```
jennwiederholen: quick question! have we decided not to use empty block period anymore?

consensus/qbft/engine/engine.go
```diff
@@ -521,12 +514,10 @@ func (e *Engine) Prepare(chain consensus.ChainHeaderReader, header *types.Header
 		firstWbftBlockNum = chain.Config().MontBlancBlock
 	}
 
-	validatorsList := validator.SortedAddresses(validators.List())
 	if firstWbftBlockNum.Cmp(header.Number) == 0 {
 		// monblac hardFork block has empty prevCommittedSeal
 		return ApplyHeaderQBFTExtra(
 			header,
-			WriteValidators(validatorsList),
```
jennwiederholen: ~nothing's done here. just return header would be fine~ > nope my mistake. guess we need it for vanityData

consensus/qbft/core/request.go
jed-wemade: Original QBFT configs are changed and `EmptyBlockPeriod` is removed.
https://github.com/wemixarchive/go-wemix-wbft/blob/d6b69aed16487bfdcdc4cb1e68248f7c93bf2bcb/README.md#modified-sturctures

Removed configs (json)
* `emptyBlockPeriodSeconds`
* `beneficiaryMode`
* `miningBeneficiary`
* `validatorSelectionMode`

core/state_transition.go
```diff
@@ -506,7 +506,26 @@ func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
 	}
 
 	// WEMIX
-	// fee is not paid to the coinbase in WEMIX
+	// In Wemix 3.0, effective tip is added to feeCollector (default: maintenance)
+	// After Montblanc hard fork, effective tip is added to coinbase same for Ethereum.
+	if st.evm.ChainConfig().IsMontBlanc(st.evm.Context.BlockNumber) {
```
egonspace: The conditional clause should be changed to when there is no `MontBlanc` hard fork or when it is in effect.

consensus/qbft/engine/engine.go
```diff
@@ -521,22 +522,228 @@ func (e *Engine) Prepare(chain consensus.ChainHeaderReader, header *types.Header
...
 	return func(qbftExtra *types.QBFTExtra) error {
-		qbftExtra.PrevRound = prevRound
-		qbftExtra.PrevPreparedSeal = prevPreparedSeal
-		qbftExtra.PrevCommittedSeal = prevCommittedSeal
+		qbftExtra.EpochInfo = epochInfo
 		return nil
 	}
 }
 
+func (e *Engine) GetEpochBlock(chain consensus.ChainHeaderReader, header *types.Header) *types.Header {
```
egonspace: Let's move this to engine_test.go.
```
func getEpochBlock(e *Engine, chain consensus.ChainHeaderReader, header *types.Header) *types.Header {
	config := chain.Config()
	for {
		header = chain.GetHeader(header.ParentHash, header.Number.Uint64()-1)
		if isEpoch, _, err := e.IsEpochBlockNumber(config, header.Number); err != nil {
			log.Crit("IsEpochBlockNumber failed", "number", header.Number, "err", err)
		} else if isEpoch {
			break
		}
	}

	return header
}
```

core/state_transition.go
jed-wemade: Fixed. Plus, related modifications for gas tip in #28 reverted.

consensus/qbft/engine/engine.go
```diff
@@ -521,22 +522,228 @@ func (e *Engine) Prepare(chain consensus.ChainHeaderReader, header *types.Header
...
+// Staker list returned must be sorted by case-insensitive order.
+func (e *Engine) GetStakers(number *big.Int, state govwbft.StateReader) []common.Address {
+	var stakers []common.Address
+	var stakerSetFromGov []common.Address
+
+	if state != nil {
+		stakerSetFromGov = govwbft.NCPStakers(state)
+	}
+
+	if len(stakerSetFromGov) <= int(e.cfg.GetConfig(number).MinStakers) {
```
egonspace, line 555: This code seems not to satisfy this rule(https://github.com/wemixarchive/go-wemix-wbft?tab=readme-ov-file#validator-selection).

consensus/qbft/engine/engine.go
```diff
@@ -885,60 +1126,51 @@ func mergeSeals(seals [][]byte, extraSeals map[common.Hash][]byte) [][]byte {
...
-	if len(vals) != len(extra.Validators) {
+	if len(vals) != len(extra.EpochInfo.Validators) {
 		return errors.New("WBFT: mismatch in ValidatorList sizes")
 	}
 
 	// Compare each element
-	for i := range vals {
-		if vals[i] != extra.Validators[i] {
+	for _, valIdx := range vals {
+		if stakers[valIdx] != extra.EpochInfo.GetValidator(valIdx) {
```
egonspace: Shouldn't it be `vals` instead of `stakers`?

consensus/qbft/engine/engine.go
```diff
@@ -521,22 +522,228 @@ func (e *Engine) Prepare(chain consensus.ChainHeaderReader, header *types.Header
...
+		log.Crit("IsEpochBlockNumber failed", "number", header.Number, "err", err)
+	} else if !isEpoch {
+		return nil
+	}
+
+	proposedSealsInEpoch := make(map[common.Address]int)
+	submittedSealsInEpoch := make(map[common.Address]int)
+	proposedCountsInEpoch := make(map[common.Address]int)
+	proposers := []common.Address{}
+	epochLength, epochHeader := 0, header
```
egonspace: This variable name(`epochHeader`) is confusing, because it is used for block iteration so it isn't epoch block.

consensus/qbft/engine/engine.go
jed-wemade: Checking both stakers and validators are needed. Should be fixed.

consensus/qbft/engine/engine.go
```diff
@@ -521,22 +522,228 @@ func (e *Engine) Prepare(chain consensus.ChainHeaderReader, header *types.Header
...
+		}
+
+		proposedSealsInEpoch[proposer] += len(prepareSigners)
+		for _, addr := range prepareSigners {
+			submittedSealsInEpoch[addr]++
+		}
+
+		// Accumulate PrevCommittedSeal counts.
+		committedSeal := extra.PrevCommittedSeal
+		commitSigners, err := e.GetSignerAddress(parent, extra.Round, committedSeal, core.SealTypeCommit)
```
egonspace: You should use `extra.PrevRound` rather than `extra.Round`.

consensus/qbft/engine/engine.go
```diff
@@ -521,20 +522,212 @@ func (e *Engine) Prepare(chain consensus.ChainHeaderReader, header *types.Header
...
+
+	if len(stakerSetFromGov) <= int(e.cfg.GetConfig(number).MinStakers) {
+		stakerSetFromConfig := e.cfg.GetConfig(number).Validators
+		stakers = append(stakers, stakerSetFromConfig...)
+	} else {
+		stakers = append(stakers, stakerSetFromGov...)
+	}
+
+	// Sort by case-insensitive.
+	sort.Slice(stakers, func(i, j int) bool { return stakers[i].Cmp(stakers[j]) < 0 })
```
egonspace: Does this need to be sorted?

consensus/qbft/engine/engine.go
jed-wemade: Changed `epochHeader` into `it`.

consensus/qbft/engine/engine.go
jed-wemade: It was a temporary constraint due to QBFT's sort policy, but now it can be removed.

consensus/qbft/engine/engine.go
jed-wemade: I might missed some points, but this is my thought.

> * Stabilization stage
>   * If current epoch is in a stabilization stage and the number of stakers is below the minimum stakers in last block of an epoch, next epoch will be in a stabilization stage.
>   * The very first validator set is defined in genesis.json, and validators in the genesis block initially have a staking amount of zero.

In stabilization stage, all validators in QBFT config are not stakers but they are regarded as stakers (excluding from staker counts).
Stakers in GovStaking contract are also valid because they'd like to become validators (including from staker counts).

> * After stabilization stage, validator selection follows below rules
>   * minimum stakers <= number of stakers <= target validators: every stakers become validators.
>   * number of stakers > target validators: validators are selected using VRF, considering staking amount and diligence.

After stabilization stage, all validators in QBFT config are excluded because they are not valid stakers anymore.
VRF feature is not designed so `TargetValidators` config is not be used.

> * number of stakers < minimum stakers: all remaining stakers become validators, which should not occur in a public network after stabilization stage for the sake of network security.

The point for this rule is to know when the stabilization stage passed away.
When the node starts consensus (or verifying) at arbitrary block, the node should traverse previous blocks to check if the chain passed away stabilization stage.
I think that the process costs high and the situation should not occur for security indeed.
For these reasons, I omitted the implementation of stabilization stage. (sorry for missing TODO comment)

Perhaps, rawdb can be used to cache the block of stabilization stage passed away, which leads rawdb messy.

jed-wemade: I missed this point.
> An epoch in a stabilization stage has the validator set which is same to the previous epoch.

Upon stabilization stage, GovStaking contract is unused?
It seems validator set is fixed of what in QBFT config.

egonspace: > Upon stabilization stage, GovStaking contract is unused?
> It seems validator set is fixed of what in QBFT config.

Yes, in stabilization stage, we should get the validators from QBFT config rather than GovStaking.

I think we need some kind of flag whether the system is in the stabilization stage or not in the `GovStaking`. With it, we can know the status referring the contract rather than traversing all chain.

consensus/qbft/engine/engine.go
```diff
@@ -606,7 +806,7 @@ func (e *Engine) ExtractGenesisValidators(header *types.Header) ([]common.Addres
 		return nil, err
 	}
 
-	return extra.Validators, nil
+	return extra.EpochInfo.GetStakers(), nil
```
egonspace: `ExtractGenesisValidators` is not used anymore. How about to remove it?

consensus/qbft/engine/engine.go
```diff
@@ -785,82 +985,116 @@ func setExtra(h *types.Header, qbftExtra *types.QBFTExtra) error {
...
+		if blockReward.Cmp(bReward) < 0 {
+			// Unreachable if genesis block is set correctly.
+			log.Crit("block reward underflow", "blockReward", blockReward, "bReward", bReward)
+		}
+		blockReward.Sub(blockReward, bReward)
+	}
+
+	getStakerInfo := func(addr common.Address) *govwbft.Staker {
+		// If the staker is removed from gov, use fallback staker info with zero staking.
+		if !govwbft.IsStaker(state, addr) {
```
egonspace, line 1030: We've decided that unstaking should left the associated staking information. This is an unexpected scenario and should therefore result in an error.

jed-wemade: Unstaking makes staking balance into zero regardless of the staking information.
Does GovStaking revert execution for accessing staker info who is unstaked?
Or Do stakers have non-zero staking balance in a short despite of unstaking?

jed-wemade: > Or Do stakers have non-zero staking balance in a short despite of unstaking?

Staking balance can be non-zero if delegators exist.

egonspace (approved):

egonspace:
TODO in other PR
- deciding validators concidering stabilization stage
- get staker info which is not a staker currently (removed in this epoch)

PR #65, opened 2025-02-06
fix: redefine round change timer and multiengine test framework

egonspace:
- redefine round change timer
  - as is:
    - ibft: start round change timer at only `startNewRound`
    - qbft: on accepting request(round 0) and at `startNewRound`(round +1) and re-start at `handlePreprepare`
  - to be:
    - wbft: start round change timer at only `startNewRound` and re-start at `handlePreprepare`
  - cause:
    - wbft engine waits for block period before making a block
    - so if we start round change timer on accepting request, engine may wait for block period time + round change timeout at most
    - if we start round change timer at `startNewRound`, block period time is included into round change timeout
    - while restarting the round change timer upon receiving a preprepare message effectively extends the round change timeout period, it is done to synchronize round changes among the validators. so we keep this policy
- multiengine test framework
  - multiengine test case


jed-wemade (approved):

egonspace:
I think we must do `newRoundChangeTimer()` in `startNewRound` like prior logic. I'll describe the reason after some thought.

jed-wemade (approved):

jed-wemade (approved):

PR #69, opened 2025-02-12
fix: prevent double round change

egonspace:
- round changing can occur twice
  - round change timeout and receiving 2/3+ timeout message can occur simultaneously
  - `startNewRound` is called by latter
  - and then timeout -> timeoutEvent sent
  - `startNewRound` ends and called again by timeoutEvent

- solution
  - during `startNewRound`, cancel prior sent timeoutEvent

### DIFF
--- a/consensus/qbft/core/core.go
+++ b/consensus/qbft/core/core.go
@@ -103,8 +103,9 @@ type Core struct {
 	currentMutex sync.Mutex
 	handlerWg    *sync.WaitGroup
 
-	roundChangeSet   *roundChangeSet
-	roundChangeTimer *time.Timer
+	roundChangeSet          *roundChangeSet
+	roundChangeTimer        *time.Timer
+	lastSentTimeoutCalceled *bool
 
 	QBFTPreparedPrepares []*qbftmessage.Prepare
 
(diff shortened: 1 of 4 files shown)


consensus/qbft/core/core.go
```diff
@@ -103,8 +103,9 @@ type Core struct {
 	currentMutex sync.Mutex
 	handlerWg    *sync.WaitGroup
 
-	roundChangeSet   *roundChangeSet
-	roundChangeTimer *time.Timer
+	roundChangeSet          *roundChangeSet
+	roundChangeTimer        *time.Timer
+	lastSentTimeoutCalceled *bool
```
jed-wemade: lastSentTimeoutCalceled -> lastSentTimeoutCanceled

event/event.go
```diff
@@ -36,8 +36,6 @@ type TypeMuxEvent struct {
 // called after mux is stopped will return ErrMuxClosed.
 //
 // The zero value is ready to use.
-//
```
jed-wemade: Should be deleted?

egonspace: I tried modify the code using `Feed` instead of `TypeMuxEvent`, but it required a lot of modification. So we must keep going with `TypeMuxEvent` for a while and I removed the `deprecated` tag from it.

jed-wemade: I agree to use `Feed` instead of `TypeMuxEvent`.
But this deprecation comment is originated from go-ethereum (although it's still used in many structs), I suggest the comment intact.

egonspace: Okay, I'll apply it.

jed-wemade (approved):

PR #70, opened 2025-02-13
fix: epoch block implementation

jed-wemade:
* Get staker list correctly and remove wrong verification.
* Generate initial epoch block correctly about WEMIX -> WBFT transition.

### DIFF
--- a/consensus/qbft/engine/engine.go
+++ b/consensus/qbft/engine/engine.go
@@ -572,15 +572,43 @@ type stakerInfo struct {
 	staker      *types.Staker
 }
 
+func (e *Engine) createInitialEpochBlock(config *params.ChainConfig, header *types.Header, state govwbft.StateReader) *types.EpochInfo {
+	var newEpoch types.EpochInfo
+
+	// Init diligence score of every staker to DefaultDiligence.
+	newStakers := e.GetStakers(config, header.Number, state)
+	newEpoch.Stakers = make([]*types.Staker, len(newStakers))
+	for i, staker := range newStakers {
+		newEpoch.Stakers[i] = &types.Staker{
+			Addr:      staker,
+			Diligence: types.DefaultDiligence,
+		}
+	}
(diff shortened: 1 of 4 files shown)


consensus/qbft/engine/engine.go
```diff
@@ -550,6 +550,10 @@ func (e *Engine) GetStakers(number *big.Int, state govwbft.StateReader) []common
 
 	if state != nil {
 		stakerSetFromGov = govwbft.NCPStakers(state)
+		// WBFT chain
```
egonspace: I think we should get stakers with different call according whether `montblanc` fork exists or not.
If it exists we get stakers from `govwbft.NCPStakers()`, otherwise we get it from `govwbft.Stakers()`.

consensus/qbft/engine/engine.go
```diff
@@ -572,17 +572,45 @@ type stakerInfo struct {
...
+}
+
 // verifyHeader() must catch inconsistent seals before calling this.
 func (e *Engine) buildEpochInfo(chain consensus.ChainHeaderReader, header *types.Header, state govwbft.StateReader) *types.EpochInfo {
 	var newEpoch types.EpochInfo
 
 	config := chain.Config()
 	if isEpoch, _, err := e.IsEpochBlockNumber(config, header.Number); err != nil {
-		log.Crit("IsEpochBlockNumber failed", "number", header.Number, "err", err)
+		log.Warn("IsEpochBlockNumber failed", "number", header.Number, "err", err)
```
egonspace: Why is it changed to `Warn`? Isn't `Crit` more suitable?

jed-wemade: I change it because `processFinalize()` catches error and for testing purposes.
If even trying to build an epoch block is disallowed for non-epoch block, this change should be reverted.

egonspace (approved):

PR #71, opened 2025-02-19
fix: missing extra and modify sync policy for wbft

egonspace:
- fix problem missing extra seals (these are removed when round changes)
- modify chain sync policy
  - before: stop miner and start sync when it is late for 1 block. ~re-sync period: 10 sec~
  - after: stop miner and start sync when it is late for 2 block for wbft. ~re-sync period: 3 sec~
  - The motivation behind this approach is that, in wBFT, being one block behind is considered a normal situation. Starting a sync process stops the engine, so this avoids unnecessarily stopping the engine. When two blocks behind, the engine cannot participate in consensus anyway, so it's preferable to stop the engine and start syncing.
- add a multiengine test


consensus/qbft/core/extraseal.go
```diff
@@ -95,27 +95,30 @@ func (c *Core) ProcessExtraSeal(lastProposal qbft.Proposal, priorRound *big.Int)
 	}
 
 	// process prepare seal
-	for _, msg := range c.prepareExtraSeals {
+	for addr, msg := range c.prepareExtraSeals {
 		if msg != nil {
 			view := msg.View()
 			if latestView.Cmp(&view) == 0 && msg.Digest == lastProposal.Hash() {
+				// this seal(c.prepareExtraSeals[addr]) is valid and re-usable for this sequence
```
jed-wemade, line 102: valid and re-usable for this sequence -> valid for merging prev seals

egonspace: "re-useable" is matter.

jed-wemade (approved):

PR #74, opened 2025-03-13
fix: assembleblock failure of TestEth2NewBlock

colinkim:
### Changes
- Modified and hardened retries logic functionality
### Why we did it
- In rare cases, the test fails because the transaction is not included in the block.
### Review requests
- Please review the code

### DIFF
--- a/eth/catalyst/api_test.go
+++ b/eth/catalyst/api_test.go
@@ -127,7 +127,15 @@ func TestEth2AssembleBlock(t *testing.T) {
 // number of transactions in it, or it has retried three times.
 func assembleWithTransactions(api *ConsensusAPI, parentHash common.Hash, params *engine.PayloadAttributes, want int) (execData *engine.ExecutableData, err error) {
 	for retries := 3; retries > 0; retries-- {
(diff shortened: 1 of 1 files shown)


jed-wemade (approved):

eth/catalyst/api_test.go
```diff
@@ -127,7 +127,15 @@ func TestEth2AssembleBlock(t *testing.T) {
 // number of transactions in it, or it has retried three times.
 func assembleWithTransactions(api *ConsensusAPI, parentHash common.Hash, params *engine.PayloadAttributes, want int) (execData *engine.ExecutableData, err error) {
 	for retries := 3; retries > 0; retries-- {
+		runnable, _ := api.eth.TxPool().Stats()
```
egonspace: The reason for the retry here is that when we asynchronously add the transaction to the transaction pool (`ethservice.TxPool().Add([]*types.Transaction{tx}, true, false)`), it takes some time for it to be promoted within the pool. Instead of checking the transaction pool, we could likely resolve this by simply adding a `Sleep(100ms)` before the continue statement on line 144.

egonspace (approved):

PR #76, opened 2025-03-19
fix:hash calculation in ReadHeaderRLP

colinkim:
####  Changes 
- Changed the internal logic of the ReadHeaderRLP function. 

#### Why we did 
- Changed block header data calculation in WBFT. 
- Quickly find block header data 

#### Review requests 
- Please review the code

### DIFF
--- a/core/rawdb/accessors_chain.go
+++ b/core/rawdb/accessors_chain.go
@@ -30,7 +30,6 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/consensus/misc/eip4844"
 	"github.com/ethereum/go-ethereum/core/types"
-	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/ethdb"
 	"github.com/ethereum/go-ethereum/log"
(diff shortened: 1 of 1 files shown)


core/rawdb/accessors_chain.go
```diff
@@ -346,9 +345,20 @@ func ReadHeaderRLP(db ethdb.Reader, hash common.Hash, number uint64) rlp.RawValu
...
+			header := new(types.Header)
+			if err := rlp.Decode(bytes.NewReader(data), header); err != nil {
+				log.Error("Invalid block header RLP", "err", err)
+				data, _ = db.Get(headerKey(number, hash))
+				return nil
+			}
+
+			if header.Hash() == hash {
+				return nil
+			}
```
egonspace:
```suggestion
			if err := rlp.Decode(bytes.NewReader(data), header); err != nil {
				log.Error("Invalid block header RLP", "err", err)
			} else if header.Hash() == hash {
				return nil
			}
```

colinkim: I wanted to make the code more explicit, but your proposed code seems more concise.
I'll take the other reviewer's comments into consideration.

egonspace (approved):

PR #89, opened 2025-04-17
fix: adjust sync condition based on TD difference

hominlee-wemade:
Adjusted the sync condition to allow synchronization if the TD difference is at least 1, when there has been no TD change during the configured forceSyncCycle period.


eth/sync.go
```diff
@@ -54,6 +54,7 @@ type chainSyncer struct {
 	force       *time.Timer
 	forced      bool // true when force timer fired
 	warned      time.Time
+	lastTD      uint64
```
egonspace:
```suggestion
	lastForcedTD      *big.Int
```

egonspace (approved):

PR #94, opened 2025-04-30
feat: add custom duration type, fix metrics server and flaky catalyst tests

hominlee-wemade:
### Summary

This PR addresses two main issues in Config.toml configuration: incorrect handling of time.Duration fields, and misconfiguration that prevented the metrics server from starting correctly.

---

### 1. feat: parse time.Duration fields using custom Duration type in config.toml

- Fields such as Recommit and NewPayloadTimeout now support string values like "1s" or "500ms" in the TOML file.
- A custom Duration type wrapping time.Duration was introduced, implementing the encoding.TextUnmarshaler and TextMarshaler interfaces to enable automatic parsing during TOML decoding.
- This ensures that time.Duration fields, which were previously ignored due to parsing issues, are now correctly applied from the configuration.

---

### 2. fix: metrics server not starting correctly when configured via config.toml

- Moved the metrics server initialization to occur after Config.toml is loaded, so that it can start correctly based on the configuration values.
- Adopted upstream PR [#30814](https://github.com/ethereum/go-ethereum/pull/30814) from go-ethereum

---

### 3. fix: eth/catalyst, trie/pathdb: fix flaky tests

- This change fixes three flaky tests `TestEth2AssembleBlock`,`TestEth2NewBlock`, `TestEth2PrepareAndGetPayload` and `TestDisable`.
- Adopted upstream PR [#29571](https://github.com/ethereum/go-ethereum/pull/29571) from go-ethereum


egonspace (approved):

jed-wemade:
> 1. feat: parse time.Duration fields from string values in config.toml

I recommend BurntSushi/toml package which works as we want.
Or, implement custom interface like [an example](https://pkg.go.dev/github.com/naoina/toml#readme-using-the-encoding-textunmarshaler-interface) in naoina/toml.

jed-wemade (approved):

egonspace (approved):
LGTM

PR #95, opened 2025-05-02
fix: translate comment chores

jennwiederholen:
Simple corrections :
1. remove korean comments 
2. change operator contract & fee recipient inteface license to MIT


colinkim:
Approve

PR #97, opened 2025-05-08
fix: missing extraSeals

eomti-wm:
- Fixed extraSeals missing during epoch change

### DIFF
--- a/consensus/qbft/core/core.go
+++ b/consensus/qbft/core/core.go
@@ -358,12 +358,12 @@ func (c *Core) newRoundChangeTimer() {
 }
 
 func (c *Core) checkValidatorSignature(data []byte, sig []byte, view qbft.View) (common.Address, error) {
-	if view.Cmp(c.currentView()) < 0 {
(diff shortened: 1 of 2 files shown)


consensus/qbft/core/core.go
```diff
@@ -357,7 +357,15 @@ func (c *Core) newRoundChangeTimer() {
...
 }
 
-func (c *Core) checkValidatorSignature(data []byte, sig []byte) (common.Address, error) {
+func (c *Core) checkValidatorSignature(data []byte, sig []byte, view qbft.View) (common.Address, error) {
+	if view.Cmp(c.currentView()) < 0 {
+		if c.state == StateAcceptRequest && view.Cmp(&qbft.View{
+			Sequence: new(big.Int).Sub(c.current.Sequence(), common.Big1),
+			Round:    c.PriorRound(),
+		}) == 0 {
+			return qbft.CheckValidatorSignature(c.priorState.validatorSet, data, sig)
```
egonspace: Shouldn't we check whether `c.priorState.validatorSet` is nil or not?

egonspace (approved):

PR #100, opened 2025-05-20
WBFT Chain Genesis Allocation Error Handling and Refactored codes.

colinkim:
### Purpose
- Some fields in govParams in Genesis.json should also be honored as default values if they are missing
- Refactoring code related to genesis Commits

### Modified
- Refactored injectContracts , getMontBlancTransition, and BuildGovTransitionParams functions to set default values if some fields related to govnarance are missing.

### Tests
- [x] If I use an address used by wbft chain in genesis.json, such as 1000, in alloc, it should output an error and not proceed.
- [x] Whether it starts correctly after copying chaindata of other node
- [x] Does node behavior after deleting ancient db work correctly?
- [x] whether there are no issues in common node restart scenarios
- [x] the dump genesis command works correctly and the genesis.json file is extracted correctly.  
- [x] Test code works without problems
- [x] config const lookup with web3 api is retrieved correctly and values are written correctly
- [x] After deleting the variables of govparam in the genesis file, does the node run normally with default config?

### DIFF
--- a/consensus/qbft/config.go
+++ b/consensus/qbft/config.go
@@ -208,26 +208,16 @@ func GetStateTransitions(chainConfig *params.ChainConfig, num *big.Int) []params
 }
 
 func getMontBlancTransition(config *params.ChainConfig) params.StateTransition {
-	st := params.StateTransition{
-		Codes: []params.CodeParam{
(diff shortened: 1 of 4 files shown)


colinkim:
Close the PR as a duplicate.

PR #101, opened 2025-05-20
fix: RLP decode error in SignedRoundChangePayload

hominlee-wemade:
### Summary

Fixed a rlp: end of list error that occurred when decoding the RoundChanges field in a Preprepare message.
```
QBFT: Error List() Signed Payload        struct=SignedRoundChangePayload err="rlp: end of list"
```

- Changed RoundChanges from []*SignedRoundChangePayload to []rlp.RawValue

- Manually decoded each item using rlp.DecodeBytes() to safely control the RLP stream


egonspace (approved):
LGTM

PR #103, opened 2025-05-21
fix: refactor chain config

egonspace:
- refactor chain config
- feat: frame work for upgrading gov contract

Additional fix([f827c94](https://github.com/wemixarchive/go-wemix-wbft/pull/103/commits/f827c943a492de7c0bed12e13a8eddc53423bc24)):
- GovStaking.afterStabilization -> Extra.EpochInfo.Stabilizing
- GovConfig.stabilizingStakerThreshold -> chainConfig.MontBlanc.WBFT.StabilizingStakersThreshold
- GovConfig.govCouncil added
- IGovCouncil added (GovNCP implements it; added emergency mode)
- add transferOperatorShip() to GovStaking.sol
- GovStaking.userInfo[staker][operator] -> GovStaking.userInfo[staker][staker]; operator address is not used for userInfo


colinkim:
There is a risk of division by zero in the code that calculates diligence.
For example, in code like this 
`d = uint64(submittedSealsInEpoch[staker]) * types.DiligenceDenominator / (2 * (epochLength - 1))`
if epochLength is 1 (the value set in genesis.json), it will return "panic: runtime error: integer divide by zero"

egonspace:
> There is a risk of division by zero in the code that calculates diligence. For example, in code like this `d = uint64(submittedSealsInEpoch[staker]) * types.DiligenceDenominator / (2 * (epochLength - 1))` if epochLength is 1 (the value set in genesis.json), it will return "panic: runtime error: integer divide by zero"

I created a new issue for this. Thank you for the comment.
https://github.com/wemixarchive/go-wemix-wbft/issues/104

consensus/qbft/types.go
```diff
@@ -270,3 +273,46 @@ type SealData struct {
...
+	return extraDataBytes, nil
+}
+
+func CreateInitialEpochInfo(config *params.MontBlancConfig) (*types.EpochInfo, error) {
+	var (
+		stakers       []common.Address
+		blsPublicKeys []string
+		epochInfo     = new(types.EpochInfo)
+	)
+	stakers = append(stakers, config.Init.Validators...)
```
colinkim: If the number of stakers and blsPublicKeys do not match, an “index out of range” panic will occur.

consensus/qbft/types.go
```diff
@@ -270,3 +273,46 @@ type SealData struct {
...
+		epochInfo.Validators = append(epochInfo.Validators, uint32(i))
+		epochInfo.BLSPublicKeys = append(epochInfo.BLSPublicKeys, hexutil.MustDecode(blsPublicKeys[i]))
+	}
+
+	log.Trace("initial epoch info", "validators", epochInfo.Validators)
+	for i, staker := range epochInfo.Stakers {
+		log.Trace(fmt.Sprintf("  - stakers[%d]", i), "addr", staker.Addr, "diligence", staker.Diligence)
+	}
+
+	return epochInfo, nil
```
colinkim: It would be nice to utilize an error that we don't currently use in conjunction with the len(config.Init.Validators) == len(config.Init.BLSPublicKeys) check.

consensus/qbft/types.go
egonspace: It is checked by `(c *MontBlancConfig) CheckValidity()`.

consensus/wemix/consensus.go
```diff
@@ -49,28 +46,28 @@ func (we *WemixConsensus) Start(
...
 
-func (we *WemixConsensus) Author(header *types.Header) (common.Address, error) {
+func (we *MontBlancConsensus) Author(header *types.Header) (common.Address, error) {
 	return header.Coinbase, nil
 }
 
-func (we *WemixConsensus) VerifyHeader(chain consensus.ChainHeaderReader, header *types.Header) error {
+func (we *MontBlancConsensus) VerifyHeader(chain consensus.ChainHeaderReader, header *types.Header) error {
 	// The fact that this engine is used implies that the MontBlanc hard fork is configured. (MontBlanc is not nil)
 	if chain.Config().IsMontBlanc(header.Number) {
```
colinkim, line 60: IMO:  What if we used the interface instead of the approach below?
```
if chain.Config().IsMontBlanc(headers[0].Number) {
		return we.wbft.VerifyHeaders(chain, headers)
} else if !chain.Config().IsMontBlanc(headers[len(headers)-1].Number) {
		return we.legacy.VerifyHeaders(chain, headers)
}
```

consensus/qbft/engine/engine.go
```diff
@@ -1183,7 +1161,7 @@ func verifyEpoch(e *Engine, chain consensus.ChainHeaderReader, header *types.Hea
 // If number of stakers <= targetValidators, use staker list as it is.
 // If number of stakers > targetValidators, random selection from the list in VRF manner
 // depending on their staking amounts and diligence score.
-func (e *Engine) decideValidators(header *types.Header, newStakers []common.Address) []uint32 {
+func (e *Engine) decideValidators(header *types.Header, newStakers []common.Address, targetValidators uint64) []uint32 {
```
jennwiederholen: This function is currently implemented to only comply with the WEMIX 3.5 policy that uses NCP. 
- The stakers parameter is always passed as the NCP list inside the getStakers function when useNCP is true.
- Thus, `targetValidators` is no need

The comment also says about wemix3.5, and wemix4.0, the concept we're not using in the codebase ( we only handles monblanc, ncp .. ) 
However, since the codebase is designed to also support chains with useNCP disabled, this function should be updated to handle the case when useNCP is false as well.
It would be good to revise the logic accordingly, something like the following:

```
if useNCP {
    // use current code
 } else {
  // is staker is less than targetValidator , use as it is, 
  // if not , randomly select validators from the stakers list based on the number specified by targetValidators (VRF-based selection can be left as a TODO for now).
  }

File: params/config_wbft.go [hunk 5]
```diff
@@ -9,35 +15,384 @@ import (
...
+			return errors.New("`montblanc.upgrades`: missing `block`")
+		}
+		if upgrade.GovContracts == nil {
+			return errors.New("`montblanc.upgrades`: missing `govContracts`")
+		}
+		if err := CheckUpgradeGovContractVersions(upgrade.GovContracts); err != nil {
+			return fmt.Errorf("`montblanc.upgrades.govContracts`: %v", err)
+		}
+	}
+
```
jennwiederholen, line 150: Missing validity check for `targetValidators.` 
According to readMe this value should be less. than or equals to epochLength
https://github.com/wemixarchive/go-wemix-wbft/blob/83f7a658894efed5fc8850c45cc25401bcef4f35/README.md?plain=1#L201

When useNCP is true, targetValidators is not used so in wemix3.5, no critical problem 
but when not using it, it can cause problem.
It is also related to the comment here i left : 
https://github.com/wemixarchive/go-wemix-wbft/pull/103/files#r2128349172

consensus/qbft/engine/engine.go
egonspace: `decideValidators` does not care whether useNCP is true or false. And it shouldn't, I think.
And I added `targetValidators` for meaning that we need this config in THIS FUNCTION and it is used for deciding validators.

params/config_wbft.go
egonspace: I applied it. Thanks.

eomti-wm (approved):

PR #107, opened 2025-06-10
feat: implement randao system

egonspace:
- add `RandaoReveal` to Extra and use `header.MixDigest` for randoMix
- add `computeShuffledIndex` shuffle function (from prysm); ethereum beacon protocol
- shuffle validators at epoch
- enable `targetValidators` config
- enable EIP-4339 (PREVRANDAO opcode)
- engine uses ecdsa key(not bls key) to sign for `RandaoReveal`
  - no need to aggregate sigs
  - bls signature is larger than ecdsa signature
  - sign data is (chainId, hard fork randao version(montBlanc = 1), block.Number)


egonspace:
This protocol may have the "RANDAO biasability" problem(https://eth2book.info/latest/part2/building_blocks/randomness/#randao-biasability).
Namely, last proposer of an epoch can "SELECT" whether he propose `randao_reveal` or not for his profit about election of next epoch validators and other validators also can "SELECT" to vote or not as same reason.

Nevertheless, we have decided not to address this further in this PR for the following reasons:
- Exploiting this vulnerability is not straightforward, as a practical attack requires a significant "tail length", which in turn necessitates a large amount of stake.
- In version 3.5, the system will operate under the NCP framework.
- This issue is planned to be resolved in the future with the implementation of a VDF (Verifiable Delay Function).

consensus/qbft/engine/engine.go
```diff
@@ -293,6 +300,19 @@ func (e *Engine) verifyCascadingFields(chain consensus.ChainHeaderReader, header
 		}
 	}
 
+	// verify extra RandaoMix and RandaoReveal
+	currentExtra, err := types.ExtractQBFTExtra(header)
```
jennwiederholen: Minor suggestion:
In the verifyHeader logic, we're currently calling the ExtractQbftExtra function multiple times unnecessarily.

We call it once at the beginning of verifyHeader (but don’t use it there), and then call it again inside verifyCascadingFields, verifySeal, and verifyPrevSeal.
This results in four separate memory allocations of qbftExtra—all with the same content—within a single verifyHeader call.

How about extracting qbftExtra once at the start of verifyHeader and passing it down to the other functions to reuse? 😃

jennwiederholen (approved):

consensus/qbft/engine/engine.go
egonspace: You're right. I applied it.

PR #109, opened 2025-06-13
feat: refactor transition and updates

jennwiederholen:
This task involves refactoring ChainConfig and qbft.Config related to the Transitions and Upgrades fields.

### Upgrades
The `Upgrades` field is a configuration for `govContract` changes. `govContracts` are built-in contracts (set in the state by code) used for the WBFT consensus.
Previously, this field was part of the `ChainConfig.Montblanc`struct. However, since the upgrade of govContracts is considered as hard fork, it is more appropriate that if hardforks following after Montblanc hardfork includes the govContract upgrades, define it in its own config.
Therefore, the Upgrades field has been removed from `ChainConfig.Montblanc` and is now managed by qbft.Config inside the WBFT config.
 
### Transitions
The `Transitions` field is a configuration for changes to the WBFT consensus settings. It used to be part of the `ChainConfig.MontblancConfig` struct. Since changes to the WBFT configuration are considered hard fork-level changes, this field has been refactored and is now managed inside the WBFT engine as well.
However, the Transitions field has not been completely removed — it can still be defined using chainConfig.Transitions for convenience, especially for private chains using WBFT.
For public chains, it is recommended to define WBFT configuration changes directly in the hard fork configurations, not via Transitions.


egonspace (approved):
LGTM

PR #110, opened 2025-06-19
fix: making our mark, finally

egonspace:
- QBFT -> WBFT
- qbftXXX -> wbftXXX
- MontBlanc -> Croissant


jennwiederholen (commented):
are we going to change montblanc -> croissant at another task?

jennwiederholen (approved):

PR #116, opened 2025-07-15
feat: add croissant logic when genesis block is already set in triedb

jennwiederholen:
The function `SetupGenesisBlockWithOverride` was missing the Croissant logic — adding extraData and allocating system contract code — when the genesis block already exists in the trie DB. So I added it.


egonspace (approved):

consensus/wbft/backend/engine_test.go
```diff
@@ -147,6 +147,7 @@ func newBlockchainFromConfig(genesis *core.Genesis, nodeKeys []*ecdsa.PrivateKey
 	// find proposer key
 	for i, key := range nodeKeys {
 		addr := crypto.PubkeyToAddress(key.PublicKey)
+		fmt.Println(addr.String())
```
egonspace: debugging code

consensus/wbft/backend/engine_test.go
```diff
@@ -109,7 +109,7 @@ func newBlockchainFromConfig(genesis *core.Genesis, nodeKeys []*ecdsa.PrivateKey
 	// Use the first key as private key
 	backend := New(cfg, nodeKeys[0], memDB)
 
-	genesis.MustCommit(memDB, triedb.NewDatabase(memDB, triedb.HashDefaults))
+	//genesis.MustCommit(memDB, triedb.NewDatabase(memDB, triedb.HashDefaults))
```
egonspace: Is this right?

jennwiederholen: Actually, it’s optional. If you don’t explicitly commit genesis, it will be set later in the `SetupGenesisBlockWithOverride()` inside `NewBlockChain()`. However, I think explicitly committing improves code readability, so I’ll revert it back :)

egonspace (approved):

PR #127, opened 2025-07-30
feat: round change retry

hominlee-wemade:
Prevents ROUND-CHANGE timeout overflow, and adds a retry mechanism to periodically re-broadcast ROUND-CHANGE messages when consensus is stalled
- The retry interval follows requestTimeoutSeconds defined in the genesis.json

### DIFF
--- a/consensus/wbft/core/core.go
+++ b/consensus/wbft/core/core.go
@@ -88,6 +88,7 @@ type Core struct {
 	events                *event.TypeMuxSubscription
 	finalCommittedSub     *event.TypeMuxSubscription
 	timeoutSub            *event.TypeMuxSubscription
+	retryTimeoutSub       *event.TypeMuxSubscription
 	futurePreprepareTimer *time.Timer
 
 	valSet     wbft.ValidatorSet
@@ -108,6 +109,8 @@ type Core struct {
 	roundChangeSet          *roundChangeSet
 	roundChangeTimer        *time.Timer
 	lastSentTimeoutCanceled *bool
+
+	retryRoundChangeTimer *time.Timer
 
 	WBFTPreparedPrepares []*wbftmessage.Prepare
 
@@ -363,6 +366,27 @@ func (c *Core) newRoundChangeTimer() {
 	*c.lastSentTimeoutCanceled = false
 	c.roundChangeTimer = time.AfterFunc(timeout, func() {
 		c.sendEvent(timeoutEvent{c.lastSentTimeoutCanceled})
+	})
+}
+
+// stopRetryTimer stops the round-change retry timer if running.
+func (c *Core) stopRetryTimer() {
+	if c.retryRoundChangeTimer != nil {
+		c.retryRoundChangeTimer.Stop()
+	}
+}
+
+// newRetryRoundChangeTimer sets a retry timer to reattempt round-change after a timeout
+func (c *Core) newRetryRoundChangeTimer() {
+	c.stopRetryTimer()
+
+	// set timeout based on the round number
+	cfg := c.config.GetConfig(c.current.Sequence())
+	timeout := time.Duration(cfg.RequestTimeout) * time.Millisecond
+
+	c.currentLogger(true, nil).Trace("WBFT: set ROUND-CHANGE retry timer", "round", c.current.Round(), "timeout", timeout.Seconds())
+	c.roundChangeTimer = time.AfterFunc(timeout, func() {
+		c.sendEvent(retryTimeoutEvent{})
 	})
 }
 
--- a/consensus/wbft/core/core.go
+++ b/consensus/wbft/core/core.go
@@ -110,7 +110,7 @@ type Core struct {
 	roundChangeTimer        *time.Timer
 	lastSentTimeoutCanceled *bool
 
-	retryRoundChangeTimer *time.Timer
+	retrySendingRoundChangeTimer *time.Timer
 
 	WBFTPreparedPrepares []*wbftmessage.Prepare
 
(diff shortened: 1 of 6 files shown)


consensus/wbft/core/core.go
```diff
@@ -354,8 +357,17 @@ func (c *Core) newRoundChangeTimer() {
 			timeout = maxRequestTimeout
 		}
 	} else {
-		// effectively impossible to observe overflow happen when maxRequestTimeout is disabled
-		timeout = baseTimeout * time.Duration(math.Pow(2, float64(round)))
+		timeoutFloat64 := math.Pow(2, float64(round)) * float64(baseTimeout)
+
+		if math.IsNaN(timeoutFloat64) || math.IsInf(timeoutFloat64, 0) || timeoutFloat64 > float64(math.MaxInt64) {
+			c.currentLogger(true, nil).Error("WBFT: Timeout overflow detected, setting timeout value to MaxInt64",
```
egonspace: The way I see it, the criteria for logging levels should be: ERROR for situations that require an operator's immediate action, and WARNING for situations that can be safely ignored for now. In this case, since no action is needed, I believe a WARNING level is sufficient.

consensus/wbft/core/roundchange.go
```diff
@@ -41,6 +41,12 @@ func (c *Core) broadcastNextRoundChange() {
 	c.broadcastRoundChange(new(big.Int).Add(cv.Round, common.Big1))
 }
 
+// broadcastRetryRoundChange re-sends ROUND-CHANGE message for the current round
+func (c *Core) broadcastRetryRoundChange() {
```
egonspace: This function doesn't actually 'resend' a round change message; it simply sends one. The context of a 'retry' exists only for the _caller_, not within the function itself. Therefore, this function appears to be a redundant wrapper and seems unnecessary.

consensus/wbft/core/core.go
```diff
@@ -366,6 +378,27 @@ func (c *Core) newRoundChangeTimer() {
 	})
 }
 
+// stopRetryTimer stops the round-change retry timer if running.
+func (c *Core) stopRetryTimer() {
```
egonspace:
```suggestion
func (c *Core) stopRetrySendingRoundChangeTimer() {
```

consensus/wbft/core/core.go
```diff
@@ -366,6 +378,27 @@ func (c *Core) newRoundChangeTimer() {
...
+	if c.retryRoundChangeTimer != nil {
+		c.retryRoundChangeTimer.Stop()
+	}
+}
+
+// newRetryRoundChangeTimer sets a retry timer to reattempt round-change after a timeout
+func (c *Core) newRetryRoundChangeTimer() {
```
egonspace:
```suggestion
func (c *Core) newRetrySendingRoundChangeTimer() {
```

consensus/wbft/core/core.go
```diff
@@ -366,6 +378,27 @@ func (c *Core) newRoundChangeTimer() {
...
+	c.stopRetryTimer()
+
+	// set timeout based on the round number
+	cfg := c.config.GetConfig(c.current.Sequence())
+	timeout := time.Duration(cfg.RequestTimeout) * time.Millisecond
+
+	c.currentLogger(true, nil).Trace("WBFT: set ROUND-CHANGE retry timer", "round", c.current.Round(), "timeout", timeout.Seconds())
+	c.roundChangeTimer = time.AfterFunc(timeout, func() {
```
egonspace: Are you sure `roundChangeTimer` is the correct timer to use here? I was wondering if you meant to use `retryRoundChangeTimer` instead.

hmlee-wm: Ah, you're right. I actually meant to use retryRoundChangeTimer.
It's surprising that the test still worked correctly. Thanks for pointing it out!

consensus/wbft/core/core.go
hmlee-wm: I agree that treating this as a WARNING is the more appropriate approach.

I originally followed Quorum's method of handling request timeouts as an ERROR, but I'll update our system to reflect this change as well.

consensus/wbft/core/roundchange.go
hmlee-wm: I removed the broadcastRetryRoundChange function and updated the logic to call broadcastRoundChange using the round value passed from the retryTimeoutEvent.

consensus/wbft/core/core.go
hmlee-wm: I applied it

consensus/wbft/core/core.go
hmlee-wm: I applied it

consensus/wbft/core/final_committed.go
```diff
@@ -20,10 +20,15 @@
...
+import (
+	"github.com/ethereum/go-ethereum/common"
+)
 
 func (c *Core) handleFinalCommittedMsg() error {
 	c.currentLogger(true, nil).Info("WBFT: handle final committed")
 	c.startNewRound(common.Big0)
+
+	// Stop ROUND-CHANGE retry timer
+	c.stopRetrySendingRoundChangeTimer()
```
egonspace: Now, we call this in `handleFinalCommittedMsg` and `handlePreprepareMsg` but I think we can call this in `stopTimer()` just once. How about this?
```
func (c *Core) stopTimer() {
	c.stopFuturePreprepareTimer()
        c.stopRetrySendingRoundChangeTimer()  // <--- in here
	if c.roundChangeTimer != nil {
		c.roundChangeTimer.Stop()
	}
	if c.lastSentTimeoutCanceled != nil {
		*c.lastSentTimeoutCanceled = true
	}
}
```

hmlee-wm: Your suggestion looks better. I'll apply it.

egonspace (approved):
LGTM

PR #128, opened 2025-07-31
fix: future msg drop

hominlee-wemade:
**Purpose**
- Add logic to drop messages whose view is ahead of the current view beyond a configured threshold, in order to improve consensus stability.

**Attack Scenario**
- An attacker sends 10,000 ROUND-CHANGE messages with increasing round numbers.
- This leads to backlog overflow and causes consensus to stall.

**Changes**
- Drop messages with a sequence or round number exceeding the configured threshold.

**Expected Effects**
- Prevents consensus disruption due to excessive future messages.
- Avoids backlog overload and memory pressure.

### DIFF
--- a/consensus/wbft/core/backlog.go
+++ b/consensus/wbft/core/backlog.go
@@ -47,13 +47,13 @@ const (
 )
 
 // isSequenceTooFarAhead returns true if the sequence difference exceeds the threshold
-func isSequenceTooFarAhead(viewSeq, currSeq *big.Int, threshold int64) (*big.Int, bool) {
+func (c *Core) isSequenceTooFarAhead(viewSeq, currSeq *big.Int, threshold int64) (*big.Int, bool) {
 	seqDiff := new(big.Int).Sub(viewSeq, currSeq)
 	return seqDiff, seqDiff.Cmp(big.NewInt(threshold)) >= 0
 }
 
 // isRoundTooFarAhead returns true if the round difference exceeds the threshold
-func isRoundTooFarAhead(viewRound, currRound *big.Int, threshold int64) (*big.Int, bool) {
+func (c *Core) isRoundTooFarAhead(viewRound, currRound *big.Int, threshold int64) (*big.Int, bool) {
 	roundDiff := new(big.Int).Sub(viewRound, currRound)
 	return roundDiff, roundDiff.Cmp(big.NewInt(threshold)) >= 0
 }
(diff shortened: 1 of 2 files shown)


0xmhha (approved):
LGTM

consensus/wbft/core/backlog.go
```diff
@@ -41,6 +41,55 @@ var (
...
+// isRoundTooFarAhead returns true if the round difference exceeds the threshold
+func (c *Core) isRoundTooFarAhead(viewRound, currRound *big.Int, threshold int64) (*big.Int, bool) {
+	roundDiff := new(big.Int).Sub(viewRound, currRound)
+	return roundDiff, roundDiff.Cmp(big.NewInt(threshold)) >= 0
+}
+
+// dropFutureTooFarMessage filters out messages too far ahead in sequence or round
+// This function prevents the node from processing messages that are excessively ahead,
+// which could disrupt the normal flow of the consensus process
+func (c *Core) dropFutureTooFarMessage(view *wbft.View) bool {
```
egonspace: The name `dropFutureTooFarMessage` is a bit misleading because it implies an action. Since the function only checks a condition, a name like `isTooFarFutureMessage` would be more accurate and clearer. What do you think?

consensus/wbft/core/backlog.go
```diff
@@ -56,6 +105,12 @@ func (c *Core) checkMessage(msgCode uint64, view *wbft.View) error {
 		return errInvalidMessage
 	}
 
+	// Drop the message if the view's sequence or round number is too far ahead
+	// This prevents processing of messages that may disrupt consensus due to excessive lead
+	if c.dropFutureTooFarMessage(view) {
+		return errFutureViewTooFar
```
egonspace, line 111: In my opinion, instead of throwing an error in this case, it would be better to just skip it and log a warning.

consensus/wbft/core/backlog.go
hmlee-wm: Yes, I agree. A name like isTooFarFutureMessage would be more accurate and clearer.

consensus/wbft/core/backlog.go
hmlee-wm: When a DDOS attack was performed, the message was handled as errFutureMessage and added to the backlog, which caused issues in system behavior.
Therefore, wouldn't it be more appropriate to return an error to prevent it from being included in the backlog?

egonspace: Ah, you're right. I was mistaken. In that case, I think it would be better to handle it this way.

egonspace (approved):

PR #130, opened 2025-08-06
fix: fee calculation logic and refactor buyGas with compatibility for WEMIX 3.0 bad blocks

hominlee-wemade:
The issue was first brought to attention in this issue: [wemixarchive/wbft#129](https://github.com/wemixarchive/wbft/issues/129)

The following two issues were fixed, fully resolving the original problem:

1. Refactored buyGas() and included the fix from PR‑29762 for EIP‑1559 balance check
- Integrated the logic from [Ethereum PR‑29762](https://github.com/ethereum/go-ethereum/pull/29762) to ensure that msg.Value is always included in the balance check when processing EIP‑1559 transactions.
- Refactored the buyGas() function to clarify fee delegation logic and reduce code duplication.

2. Fixed incorrect fee calculation logic for feePayer transactions in WEMIX 3.0
- Addressed the issue where GasFeeCap was ignored and GasPrice was incorrectly used for fee calculation in legacy WEMIX 3.0 transactions.
- For blocks before the Croissant hard fork, the legacy WEMIX 3.0 fee calculation logic (despite being incorrect) is retained to ensure compatibility with historical bad blocks.

### DIFF
--- a/cmd/evm/testdata/12/readme.md
+++ b/cmd/evm/testdata/12/readme.md
@@ -30,7 +30,7 @@ INFO [03-09|10:43:12.650] Trie dumping complete                    accounts=1 el
     "rejected": [
       {
         "index": 0,
-        "error": "insufficient funds for gas * price + value: address 0xa94f5374Fce5edBC8E2a8697C15331677e6EbF0B have 84000000 want 84000032"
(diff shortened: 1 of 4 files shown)


egonspace (approved):
LGTM

PR #146, opened 2025-09-04
fix: remove extra info cache

egonspace:
`epochCache` may store epoch info of a non-canonical block(this can be just candidate proposal).


0xmhha (approved):
LGTM

colinkim:
Changing it to something like your suggestion seems like a radical solution, but it would be expensive. 
How about incremental improvements, such as 1+2 proposals, then monitoring, then 1+3 proposals, then monitoring, etc. 
For example 
1. using a composite key instead of a block number as a key 
  Cache Improvement Proposal
`type EpochCacheKey struct {
      Number uint64
      Hash   common.Hash
  }`
3. 1an+inmemoryCache Incremental

PR #153, opened 2025-09-08
fix: missing extraSeals on round change

eomti-wm:
- Fixed an issue where `preparedSeals` for block `n` overwrote the `preparedSeals (extraSeals)` for block `n-1` during a round change in prepare.


egonspace (approved):
LGTM

PR #155, opened 2025-09-12
fix: adjust default priority fee to 100Gwei

egonspace:
Modify default priority fee per gas from 1 Gwei to 100 Gwei for compatibility with Wemix3.0.

### DIFF
--- a/miner/miner.go
+++ b/miner/miner.go
@@ -60,10 +60,20 @@ type Config struct {
 	SimulatedEnabled bool `toml:",omitempty"`
 }
 
+const (
+	// DefaultGasPriceGWei is the default gas price for Wemix 3.0 compatibility.
+	DefaultGasPriceGWei = 100
+)
+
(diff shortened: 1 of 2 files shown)


miner/miner.go
colinkim, line 1:
```
suggestion

// Wemix miner gas price constants
const (
    // DefaultGasPriceGWei is the default gas price for Wemix 3.0 compatibility.
    DefaultGasPriceGWei = 100
)

// Pre-computed gas price values
var (
    DefaultGasPrice = big.NewInt(DefaultGasPriceGWei * params.GWei)
)

// DefaultConfig contains default settings for miner.
var DefaultConfig = Config{
    GasCeil:  105000000,
    GasPrice: DefaultGasPrice, // Use pre-computed constant for clarity and maintenance

    // ... (rest of the fields) ...
}

hominlee-wemade (approved):
LGTM

0xmhha (approved):
LGTM

PR #156, opened 2025-09-17
fix: update dev-test-work.yaml

colinkim:
fixed wbft-script sync


egonspace:
I already applied it at https://github.com/wemixarchive/go-wemix-wbft/pull/152.

PR #157, opened 2025-09-26
fix: define croissant block temporarily

egonspace:
Since the Testnet chain's block height already exceeds 100,000,000, the chain sync process keeps failing. We're temporarily changing the block limit to 200,000,000 to fix this.


hominlee-wemade (approved):
LGTM

PR #158, opened 2025-09-26
fix: stable version

egonspace:
modify version meta to 'stable'.


colinkim (approved):
Congrats on reaching the stable milestone! Looking forward to what’s next

PR #164, opened 2025-10-28
fix: modify readme for general wbft

egonspace:
- modify readme
- fix dev-release workflow to include all binaries to tar


colinkim:
The phrase “4.x” appears in several places in the ReadME.md. Shouldn't it be updated accordingly?

egonspace:
> The phrase “4.x” appears in several places in the ReadME.md. Shouldn't it be updated accordingly?

applied it.

colinkim (approved):

PR #167, opened 2025-11-25
feat: support eth_GetReceiptsByHash

Wemade-Micah:
- Modified to support the RPC method eth_getReceiptsByHash, which is already supported in Wemix 3.0. 
- Since it performs the same logic as eth_getBlockReceipts, it is implemented to internally call the GetBlockReceipts function.

### DIFF
--- a/internal/ethapi/api.go
+++ b/internal/ethapi/api.go
@@ -1016,6 +1016,10 @@ func (s *BlockChainAPI) GetBlockReceipts(ctx context.Context, blockNrOrHash rpc.
 	return result, nil
 }
 
+func (s *BlockChainAPI) GetReceiptsByHash(ctx context.Context, blockHash common.Hash) ([]map[string]interface{}, error) {
+	return s.GetBlockReceipts(ctx, rpc.BlockNumberOrHashWithHash(blockHash, true))
+}
+
 // OverrideAccount indicates the overriding fields of account during the execution
 // of a message call.
 // Note, state and stateDiff can't be specified at the same time. If state is


egonspace (approved):

egonspace:
Congratulations on your first contribution from outside the team!

hominlee-wemade:
The RPC method (`eth_getReceiptsByHash`) is correctly implemented, but it's currently not exposed to the console environment. 
As a result, calling `eth.getReceiptsByHash()` via attach is not supported.

It would be good to also register the method in `web3ext.go` under the `EthJs` section, similar to the example below:
```
new web3._extend.Method({
    name: 'getReceiptsByHash',
    call: 'eth_getReceiptsByHash',
    params: 1,
    outputFormatter: function(receipts) {
        var formatted = [];
        for (var i = 0; i < receipts.length; i++) {
            formatted.push(web3._extend.formatters.outputTransactionReceiptFormatter(receipts[i]));
        }
        return formatted;
    }
}),
```

egonspace:
> The RPC method (`eth_getReceiptsByHash`) is correctly implemented, but it's currently not exposed to the console environment. As a result, calling `eth.getReceiptsByHash()` via attach is not supported.
> 
> It would be good to also register the method in `web3ext.go` under the `EthJs` section, similar to the example below:
> 
> ```
> new web3._extend.Method({
>     name: 'getReceiptsByHash',
>     call: 'eth_getReceiptsByHash',
>     params: 1,
>     outputFormatter: function(receipts) {
>         var formatted = [];
>         for (var i = 0; i < receipts.length; i++) {
>             formatted.push(web3._extend.formatters.outputTransactionReceiptFormatter(receipts[i]));
>         }
>         return formatted;
>     }
> }),
> ```

Breaking backward compatibility in the console shouldn't be a big issue. The current priority is to enable RPC usage from external servers.

eomti-wm (approved):

PR #170, opened 2026-03-04
feat: EIP-7702 Set Code Transaction support

eomti-wm:
Port EIP-7702 (Set Code Transaction) from [go-stablenet#42](https://github.com/stable-net/go-stablenet/pull/42) with gas estimation fix from [go-ethereum#33849](https://github.com/ethereum/go-ethereum/pull/33849).

EIP-7702 allows EOAs to delegate code execution to smart contracts via SetCode authorizations, enabling account abstraction. Gated behind the **Croissant hard fork**.

### Changes

- **Transaction Type**: Add `SetCodeTx` (type `0x04`) with `SetCodeAuthorization`, `croissantSigner`, and `authorizationList` RPC support. Refactor `sigHash` into inner tx type methods.
- **Execution**: Resolve delegation prefix `0xef0100` in EVM via `resolveCode`/`resolveCodeHash`. Apply authorizations during state transition. Allow delegated accounts to send transactions.
- **Transaction Pool**: Gate `SetCodeTxType` behind `IsCroissant()`. Track authorities and limit delegated accounts to one in-flight tx. Extract reservation logic to `reserver.go`.
- **Tracing**: Update tracer interface (`CaptureStart`/`CaptureTxStart` signature change). Prestate tracer captures authority state before authorization and resolves delegation targets.
- **Gas Estimation**: Include `AuthorizationList` in `callArgs` during `setDefaults` to prevent incorrect gas estimates for SetCode transactions.

### References
- [EIP-7702](https://eips.ethereum.org/EIPS/eip-7702)
- [go-stablenet#42](https://github.com/stable-net/go-stablenet/pull/42)
- [go-ethereum#33849](https://github.com/ethereum/go-ethereum/pull/33849)


internal/ethapi/transaction_args.go
```diff
@@ -466,6 +471,27 @@ func (args *TransactionArgs) ToMessage(globalGasCap uint64, baseFee *big.Int) (*
...
 	switch {
+	case args.AuthorizationList != nil:
+		al := types.AccessList{}
+		if args.AccessList != nil {
+			al = *args.AccessList
+		}
+		authList := []types.SetCodeAuthorization{}
+		if args.AuthorizationList != nil {
+			authList = args.AuthorizationList
+		}
```
0xmhha: `args.AuthorizationList != nil` is already guaranteed true by the enclosing
  `case` on line 474, so the inner `if args.AuthorizationList != nil` on line 480
  is redundant. This can be simplified to:

      authList := args.AuthorizationList

  Not a bug, but a minor cleanup.

eomti-wm: Fixed. Removed the redundant nil check and assigned directly. Thanks!

0xmhha (approved):
LGTM

code0xff (approved):
LGTM

PR #174, opened 2026-03-09
fix: config initialization and parsing improvements

eomti-wm:
## Summary

Fix chain config initialization and comma-separated config string parsing issues.

### 1. Normalize comma-separated NCP config string

- Use `splitAndTrim` instead of raw `strings.Split` for NCP address parsing during system contract initialization
- Trims whitespace and drops empty entries to prevent zero-address registration
- Returns error if `govNCP` is configured but no initial NCPs are provided

### 2. Resolve chainconfig engine mismatch

- Add `LoadChainConfigWithOverride` to resolve the final `ChainConfig` with overrides applied before engine creation
- Previously, the WBFT engine was initialized with a stale `ChainConfig` from the database, before overrides were applied in `NewBlockChain`, causing a mismatch
- Extract `validateCroissantGenesisConfig` and `initializeCroissantGenesis` helper functions
- Move override construction before engine creation in `eth/backend.go`
- Add `TestLoadChainConfigWithOverride`

### DIFF
--- a/cmd/utils/flags.go
+++ b/cmd/utils/flags.go
@@ -2097,7 +2097,18 @@ func MakeChain(ctx *cli.Context, stack *node.Node, readonly bool) (*core.BlockCh
 		gspec   = MakeGenesis(ctx)
 		chainDb = MakeChainDatabase(ctx, stack, readonly)
 	)
-	config, err := core.LoadChainConfig(chainDb, gspec)
+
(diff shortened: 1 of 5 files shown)


hominlee-wemade (approved):
LGTM

0xmhha (approved):
LGTM

PR #182, opened 2026-04-16
fix: resolve data races in wbft consensus core

eomti-wm:
### Summary

This PR fixes data races in the WBFT consensus engine identified by the Go race detector (`-race`), focusing on production runtime concurrency issues.

### Problem

The Go race detector reported the following concurrent read/write conflicts at runtime:

**`stopTimer()` concurrent calls** (`core.go`): The consensus event loop goroutine and `Core.Stop()` goroutine called `stopTimer()` simultaneously, causing a race on timer fields (`roundChangeTimer`, `lastSentTimeoutCanceled`, `futurePreprepareTimer`).

**`newRoundChangeTimer()` unsynchronized reads of `c.current`** (`core.go`, `preprepare.go`): During node startup, `startNewRound()` runs on the `Start()` goroutine while `handleEvents()` may concurrently receive a PRE-PREPARE message and invoke `newRoundChangeTimer()` through `handlePreprepareMsg()`. The function previously read `c.current` (nil-wait loop, `Sequence()`, `Round()`) without holding `currentMutex`, creating an unsynchronized read/write that the race detector flags and that can produce an incorrect timer duration.

#### `consensus/wbft/core/core.go`
- **Add `timerMu sync.Mutex`**: Protects concurrent access to timer fields (`roundChangeTimer`, `lastSentTimeoutCanceled`, `futurePreprepareTimer`) accessed simultaneously from the consensus event loop and `Core.Stop()`.
- **`currentMutex sync.Mutex` → `sync.RWMutex`**: Allows safe concurrent reads of current round state; only mutating paths take a write lock.
- **`newRoundChangeTimer()` signature change to `newRoundChangeTimer(seq, round *big.Int)`**: Eliminates unsynchronized reads of `c.current` inside the function. Callers are required to snapshot `seq` and `round` under `currentMutex` before calling. `startNewRound()` passes values directly since it already holds `Lock()`; `handlePreprepareMsg()` snapshots under `RLock()` and releases before the call. Mirrors the existing snapshot contract of `newRetrySendingRoundChangeTimer`.

#### `consensus/wbft/core/preprepare.go`
- **Snapshot `c.current` under `currentMutex.RLock()` before calling `newRoundChangeTimer`**: Synchronizes the read of `c.current.Sequence()` and `c.current.Round()` against concurrent writes in `startNewRound()`. The lock is released before the call to avoid a potential deadlock with the write lock held by `startNewRound()`.

### DIFF
--- a/consensus/wbft/core/core.go
+++ b/consensus/wbft/core/core.go
@@ -103,12 +103,13 @@ type Core struct {
 	priorState        priorState
 
 	current      *roundState
-	currentMutex sync.Mutex
+	currentMutex sync.RWMutex
 	handlerWg    *sync.WaitGroup
 
 	roundChangeSet          *roundChangeSet
 	roundChangeTimer        *time.Timer
 	lastSentTimeoutCanceled *bool
+	timerMu                 sync.Mutex // protects roundChangeTimer, lastSentTimeoutCanceled, and timer stop/start
 
 	retrySendingRoundChangeTimer *time.Timer
 
@@ -151,6 +152,8 @@ func (c *Core) GetProposer() common.Address {
 }
 
 func (c *Core) IsCurrentProposal(blockHash common.Hash) bool {
+	c.currentMutex.RLock()
+	defer c.currentMutex.RUnlock()
 	return c.current != nil && c.current.pendingRequest != nil && c.current.pendingRequest.Proposal.Hash() == blockHash
 }
 
@@ -311,19 +314,27 @@ func (c *Core) stopFuturePreprepareTimer() {
 		c.futurePreprepareTimer.Stop()
 	}
 }
-
+func (c *Core) stopRoundChangeTimer() {
+	if c.roundChangeTimer != nil {
+		c.roundChangeTimer.Stop()
+	}
+}
+func (c *Core) cancelLastSentTimeout() {
+	if c.lastSentTimeoutCanceled != nil {
+		*c.lastSentTimeoutCanceled = true
+	}
+}
 func (c *Core) stopTimer() {
+	c.timerMu.Lock()
+	defer c.timerMu.Unlock()
 	c.stopFuturePreprepareTimer()
 
 	// Stop retry sending ROUND-CHANGE retry timer
 	c.stopRetrySendingRoundChangeTimer()
 
-	if c.roundChangeTimer != nil {
-		c.roundChangeTimer.Stop()
-	}
-	if c.lastSentTimeoutCanceled != nil {
-		*c.lastSentTimeoutCanceled = true
-	}
+	c.stopRoundChangeTimer()
+
+	c.cancelLastSentTimeout()
 }
 
 func (c *Core) newRoundChangeTimer() {
@@ -375,11 +386,13 @@ func (c *Core) newRoundChangeTimer() {
 	}
 
 	c.currentLogger(true, nil).Trace("WBFT: start new ROUND-CHANGE timer", "timeout", timeout.Seconds())
+	c.timerMu.Lock()
 	c.lastSentTimeoutCanceled = new(bool)
 	*c.lastSentTimeoutCanceled = false
 	c.roundChangeTimer = time.AfterFunc(timeout, func() {
 		c.sendEvent(timeoutEvent{c.lastSentTimeoutCanceled})
 	})
+	c.timerMu.Unlock()
 }
 
 // stopRetrySendingRoundChangeTimer stops the round-change retry timer if running.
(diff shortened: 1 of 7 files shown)


hominlee-wemade (approved):

consensus/wbft/core/core.go
```diff
@@ -375,11 +386,13 @@ func (c *Core) newRoundChangeTimer() {
 	}
 
 	c.currentLogger(true, nil).Trace("WBFT: start new ROUND-CHANGE timer", "timeout", timeout.Seconds())
+	c.timerMu.Lock()
 	c.lastSentTimeoutCanceled = new(bool)
 	*c.lastSentTimeoutCanceled = false
 	c.roundChangeTimer = time.AfterFunc(timeout, func() {
 		c.sendEvent(timeoutEvent{c.lastSentTimeoutCanceled})
 	})
```
colinkim:
```suggestion
        canceled := new(bool)
        c.lastSentTimeoutCanceled = canceled
        c.roundChangeTimer = time.AfterFunc(timeout, func() {
                c.sendEvent(timeoutEvent{canceled})
        })
  ```

eomti-wm: fix : 8ff8f38377f3e534913b4371549d240077c71d00

colinkim (approved):
LGTM

0xmhha (commented):
Inside `newRoundChangeTimer`, three reads of `c.current` were performed **without mutex protection**:

  - **`core.go:343`** — `for c.current == nil` spin-wait loop (busy-loop that burns CPU until the value becomes non-nil). The loop condition itself is an
  unprotected read of shared state.
  - **`core.go:348`** — `c.current.Sequence()` call. Dereferences `c.current` without a lock.
  - **`core.go:350`** — `c.current.Round()` call. Same, unprotected.

  #### Caller audit

  This function has two call sites. Checking whether `currentMutex` is held at each:

  - **`core.go:262` `startNewRound → newRoundChangeTimer`**: enters with `currentMutex.Lock()` already held. ✓ Safe.
  - **`preprepare.go:169` `handlePreprepareMsg → newRoundChangeTimer`**: no lock held at all. ✗

  So on the `preprepare.go:169` path, the `c.current` reads inside `newRoundChangeTimer` are defenseless. If `startNewRound` writes to `c.current` at the
  same moment (`core.go:278` / `core.go:284`), read and write collide — that is the race condition (two goroutines touching the same variable concurrently,
  each breaking the other's behavior).

  #### Reproducibility

  The window in which the race can actually fire is narrow. Concretely:

  - In `handler.go:40-45`, `Start()` spawns the event goroutine with `go c.handleEvents()` and then **immediately** calls `c.startNewRound(common.Big0)`.
  - For that brief moment two goroutines run concurrently: the newly spawned `handleEvents` and the main flow's `startNewRound`. If a `PRE-PREPARE` message
  arrives from the network during this sliver, `handlePreprepareMsg → newRoundChangeTimer` is reading `c.current` while `startNewRound` overwrites it — race
   established.
  - Once the node reaches steady state, every in-loop path (`handleTimeoutMsg`, `handlePreprepareMsg`, etc.) is serialized on a **single goroutine**
  (`handleEvents`), so they cannot race against each other.

  #### But `-race` builds don't care how narrow the window is

  No matter how low the real-world collision probability, `go build -race` (Go's data-race detector build flag) checks one thing only: **"does an
  unprotected read/write pair exist on the same variable?"** Whether the window is μs-wide, or almost never triggered during a real boot sequence, the
  pattern itself gets flagged. So even if operational frequency is low, the fix is required — as long as CI runs `-race`, the build will break.

eomti-wm:
> Inside `newRoundChangeTimer`, three reads of `c.current` were performed **without mutex protection**:
> 
> * **`core.go:343`** — `for c.current == nil` spin-wait loop (busy-loop that burns CPU until the value becomes non-nil). The loop condition itself is an
>   unprotected read of shared state.
> * **`core.go:348`** — `c.current.Sequence()` call. Dereferences `c.current` without a lock.
> * **`core.go:350`** — `c.current.Round()` call. Same, unprotected.
> 
> #### Caller audit
> This function has two call sites. Checking whether `currentMutex` is held at each:
> 
> * **`core.go:262` `startNewRound → newRoundChangeTimer`**: enters with `currentMutex.Lock()` already held. ✓ Safe.
> * **`preprepare.go:169` `handlePreprepareMsg → newRoundChangeTimer`**: no lock held at all. ✗
> 
> So on the `preprepare.go:169` path, the `c.current` reads inside `newRoundChangeTimer` are defenseless. If `startNewRound` writes to `c.current` at the same moment (`core.go:278` / `core.go:284`), read and write collide — that is the race condition (two goroutines touching the same variable concurrently, each breaking the other's behavior).
> 
> #### Reproducibility
> The window in which the race can actually fire is narrow. Concretely:
> 
> * In `handler.go:40-45`, `Start()` spawns the event goroutine with `go c.handleEvents()` and then **immediately** calls `c.startNewRound(common.Big0)`.
> * For that brief moment two goroutines run concurrently: the newly spawned `handleEvents` and the main flow's `startNewRound`. If a `PRE-PREPARE` message
>   arrives from the network during this sliver, `handlePreprepareMsg → newRoundChangeTimer` is reading `c.current` while `startNewRound` overwrites it — race
>   established.
> * Once the node reaches steady state, every in-loop path (`handleTimeoutMsg`, `handlePreprepareMsg`, etc.) is serialized on a **single goroutine**
>   (`handleEvents`), so they cannot race against each other.
> 
> #### But `-race` builds don't care how narrow the window is
> No matter how low the real-world collision probability, `go build -race` (Go's data-race detector build flag) checks one thing only: **"does an unprotected read/write pair exist on the same variable?"** Whether the window is μs-wide, or almost never triggered during a real boot sequence, the pattern itself gets flagged. So even if operational frequency is low, the fix is required — as long as CI runs `-race`, the build will break.

fix : c2de503629d620f0f4984a035b4bc6b4c801cf4b

0xmhha (approved):
LGTM

PR #184, opened 2026-04-20
fix(govstaking): scope operator claim and refactor withdraw for safety and liveness

0xmhha:
## Summary

Fixes two classes of high-severity issues in the GovStaking system contract and restores liveness for users holding credentials with mixed unbonding periods.

- **Operator permission scope on `claim()`**: any registered operator could claim rewards from any staker in the system, not just the staker they operate.
- **Early drain on `withdraw()`**: multi-credential withdrawals only validated the last credential's maturity, allowing earlier, still-locked credentials to be pulled through the bulk path whenever the credential sequence was not monotonically ordered by expiry.
- **Withdraw liveness**: the original sequential-lock policy trapped already-mature credentials behind an earlier, still-locked one. `withdraw()` now skips locked/empty slots and drains every mature credential it encounters, while preserving the per-credential expiry check.

Includes updated Solidity source, regenerated bytecode (embedded into the `gwemix` binary via `go:embed`), contract-level NatSpec explaining the Staker/Operator role model, and a dedicated regression + unit test suite.

### Issue 1 — Operator permission scope on `claim()`

The condition inside `claim()` used a global operator check:

```solidity
address _user = isOperator(msg.sender) ? _staker : msg.sender;

function isOperator(address _addr) public view returns (bool) {
    return stakerByOperator[_addr] != address(0);
}
```

`isOperator(msg.sender)` returns `true` for any address registered as an operator of any staker. Once `_user` was set to `_staker`, the caller could drain the self-staking reward accumulated against that staker regardless of whether they actually operate it.

Attack shape:
1. Operator A is registered as the operator of Staker X (`stakerByOperator[A] = X`).
2. Staker Y accumulates a self-staking reward in `userRewardInfo[Y][Y].pendingReward`.
3. A calls `claim(Y, false)`. The condition evaluates to `true`, `_user = Y`, and Y's reward is sent to A.
4. Repeatable — any operator can drain any staker's self-staking reward.

Scope: any registered operator in the network. Impact: complete loss of self-staking rewards across all stakers.

### Issue 2 — Early drain on `withdraw()` with non-monotonic credentials

`withdraw(_withdrawalCount > 0)` asserted maturity only on the last credential in the requested range:

```solidity
require(
    credentials[msg.sender][withdrawalIndex + _withdrawalCount - 1].withdrawableTime
        <= block.timestamp,
    "withdrawal time not reached"
);
```

The loop performed no per-credential maturity check. This was safe only while `withdrawableTime` was monotonically increasing with credential index. Two conditions break that assumption:

**Trigger A — Governance parameter change.**
Shortening `unbondingPeriodStaker` or `unbondingPeriodDelegator` while earlier credentials are still in-flight produces a later credential that matures earlier than an existing one. A small follow-up unstake suffices to pull a much larger, still-locked credential through the bulk path.

```
T=0       unstake 1000        -> credential[0].withdrawableTime = T + 28 days
T=1 day   [governance]        -> unbondingPeriodStaker: 28 days -> 2 days
T=1 day   unstake 1 wei       -> credential[1].withdrawableTime = T + 3 days
T=3 days  withdraw(2)         last-credential check passes (credential[1] matures at T+3d)
                              loop has no per-item check
                              -> credential[0] (1000) drained 25 days early
```

**Trigger B — Single account holding mixed credential types.**
An account can be the operator of one staker and a delegator to another at the same time:
- `unstake()` creates credentials with `unbondingPeriodStaker`.
- `undelegate()` creates credentials with `unbondingPeriodDelegator`.

In typical deployments `unbondingPeriodStaker > unbondingPeriodDelegator`, so a subsequent undelegation produces a credential that matures before an earlier unstake credential — triggering the same bypass **without any governance change**. This path is reproducible in default configurations whenever the same address holds both roles.

```
unbondingPeriodStaker = 28 days, unbondingPeriodDelegator = 7 days
T=0       OperatorA.unstake(1000)       -> credentials[A][0].withdrawableTime = T + 28d
T=1 day   OperatorA.undelegate(Y, 100)  -> credentials[A][1].withdrawableTime = T + 8d
T=8 days  OperatorA.withdraw(2)         -> credentials[A][0] (1000) drained 20 days early
```

### Issue 3 — Liveness: mature credentials trapped behind locked ones

Even with per-credential maturity enforcement, the original sequential-lock policy made an already-mature credential unreachable whenever an earlier credential remained locked:

- Auto mode (`count == 0`) broke at the first not-yet-mature credential.
- Explicit mode (`count > 0`) reverted on any locked credential inside the requested range.

The same mixed-period condition that triggered Issue 2 therefore also caused a liveness problem for legitimate users: the short-period credential's funds remained inaccessible until the long-period credential matured, even though each credential is an independent withdrawal obligation.

### `claim()` scope — bind operator to a specific staker

```diff
 function claim(address _staker, bool _restake)
     external isRegistered(_staker)
     inspectWithCouncil(GovStaking.claim.selector, abi.encode(_staker, _restake))
 {
-    address _user = isOperator(msg.sender) ? _staker : msg.sender;
+    // Operator may claim on behalf of a staker only when msg.sender is the
+    // operator registered for that specific staker, not any staker in the
+    // system.
+    address _user = (stakerByOperator[msg.sender] == _staker)
+                    ? _staker : msg.sender;
```

Legitimate paths preserved:

| Caller's role vs `_staker` | Result |
| --- | --- |
| Operator of `_staker` | `_user = _staker` (unchanged) |
| Operator of another staker | `_user = msg.sender` → existing "no reward to claim" guard fires |
| Plain delegator self-claim | `_user = msg.sender` (unchanged) |
| Operator acting as a delegator on a different staker | `_user = msg.sender` → claims their own delegation reward (unchanged) |

`isOperator(...)` public view stays as-is (used by off-chain queries); only the internal scope check in `claim()` is tightened.

### `withdraw()` — per-credential maturity and liveness

`withdraw()` is refactored so that the loop scans every credential in the range and:

- **Skips empty slots** (already withdrawn via an earlier call).
- **Skips locked credentials** (still-maturing) — they are never drained.
- **Drains mature credentials** one at a time using checks-effects-interactions.
- **Advances `withdrawalIndex` to the oldest still-live slot** at the end of the scan, folding consecutive empty-prefix slots into the advance so subsequent calls start at the first live slot.

Mode semantics after the refactor:

| `_withdrawalCount` | Behavior |
| --- | --- |
| `0` (auto) | Drain every currently-mature credential. |
| `N > 0` (explicit) | Drain exactly `N` mature credentials in order. If the scan cannot find `N` matures, the whole call reverts with `insufficient withdrawable credentials`. |

`Withdrawn` now emits the actual storage index of the withdrawn credential.

Complete behavior matrix:

| State | Before | After |
| --- | --- | --- |
| All mature, `count == 0` | drain all | drain all (unchanged) |
| All mature, `count > 0` | drain all | drain exactly `count` |
| Monotonic, partial mature, `count == 0` | break at first locked | break at first locked (unchanged) |
| Monotonic, partial mature, `count > 0` covering mature range | succeed | succeed (unchanged) |
| Non-monotonic, `count > 0` including a locked credential | **silently drain locked** | scan skips locked; if scan cannot find `count` matures, revert with `insufficient withdrawable credentials` |
| Non-monotonic, `count == 0` | break at first locked — **mature credentials behind it unreachable** | **drain every mature credential, skip locked** |
| `count > available matures` | `out of max user credential index` revert | `insufficient withdrawable credentials` revert |
| Empty credential set | `no credential to withdraw` revert | `no credential to withdraw` revert (unchanged) |

### Documentation — Staker/Operator role model

A contract-level NatSpec block clarifies the roles:

- **Staker** is the on-chain identity of a Validator. The address registered as the `staker` is the same node-key address the consensus layer uses to sign blocks.
- **Operator** is a separate EOA that acts on behalf of the Staker for staking and reward-management interactions, so the validator node key does not have to be exposed to routine wallet operations.

## Hardfork compatibility (go-wemix → go-wbft)

Only function bodies change. All external surfaces remain constant:

- Storage slots `0x0` through `0xe` are unchanged.
- Function signatures (`claim(address,bool)`, `withdraw(uint256)`, and all others) are unchanged.
- Event signatures are unchanged. Only the semantic meaning of `Withdrawn.withdrawalIndex` changes from a post-increment counter to the actual storage key of the withdrawn credential.
- State injected by `GetGovContractsTransition()` (BLS PoP address, `govConfig`, `govRewardeeImp`) is unchanged.
- `contracts.go` embed paths and the `GovContractCodes` mapping are unchanged.
- Go bindings in `wemixgov/bind/` are unaffected.

The `v1` source is the authoritative target: the bytecode is deployed for the first time at the Croissant fork block via `InjectContracts()`, so the updated bytecode is the one users see. The placeholder at `contracts-wbft/v2/GovStaking.sol` stays a placeholder; no upgrade is introduced by this change.

## Files changed

- `wemixgov/governance-contract/contracts-wbft/v1/GovStaking.sol` — `claim()` scope tightening, `withdraw()` refactor, contract-level NatSpec role model.
- `wemixgov/governance-wbft/govcontracts/v1/GovStaking` — regenerated runtime bytecode (solc 0.8.14 via `compile.go`).
- `wemixgov/governance-contract/test/gov_security_test.go` — new integration tests covering claim scope, safety under governance-shortened unbonding, safety + liveness on mixed staker/delegator credentials, pointer advancement, partial-mature drain, insufficient-count revert, and the `receive()` gate.
- `wemixgov/governance-contract/test/gov_wbft_test.go` — revert-message expectations aligned with the new explicit-mode revert.
- `wemixgov/governance-wbft/stateutil_test.go` — slot math, `EnumerableSet`, short/long/empty dynamic bytes.
- `wemixgov/governance-wbft/governance_test.go` — `GetGovContractsTransition` full image, govCouncil injection, invalid params, NCP empty-list guard, version whitelist, `splitAndTrim`.
- `wemixgov/governance-wbft/ncp_test.go` — `initializeNCP` unique / duplicated / empty, `NCPAt`.

Unrelated test fixture updates inherited from the base branch (regenerated to track current deterministic output): `cmd/devp2p/internal/ethtest/testdata/*`, `internal/ethapi/testdata/*`.

## Verification

```bash
go build ./...
go test -timeout 900s -count=1 ./wemixgov/...
```

- `wemixgov/governance-contract/test` — ~42s, includes the new safety and liveness regression cases alongside all pre-existing integration tests.
- `wemixgov/governance-wbft` — <1s, new unit tests covering the storage layer helpers and governance transition builder.

## Test plan

- [x] `go test -timeout 900s -count=1 ./wemixgov/...`
- [x] `cd wemixgov/governance-contract/contracts-wbft && go run compile.go` produces no diff against the committed `wemixgov/governance-wbft/govcontracts/v1/GovStaking`.
- [x] `go build ./...`


wemixgov/governance-contract/contracts-wbft/v1/GovStaking.sol
```diff
@@ -509,29 +536,80 @@ contract GovStaking {
...
-                break;
+
+            // Explicit mode: once the requested count is met, keep scanning
+            // only to determine the correct pointer advance target.
+            if (!_autoMode && _processed == _withdrawalCount) {
+                if (_oldestAlive == type(uint256).max && _credential.withdrawableTime > 0) {
+                    _oldestAlive = i;
+                    break;
+                }
+                continue;
```
eomti-wm, line 584: In explicit mode, once `_processed == _withdrawalCount` is reached and `_oldestAlive` is already set from a previously encountered credential that has not yet reached its withdrawable time, the inner if condition fails and the loop continues iterating over all remaining slots unnecessarily, incurring redundant storage reads.

0xmhha: fix : 066708184c07d84c720f2da12cb2c1560049e918

eomti-wm (approved):
LGTM

colinkim (approved):
LGTM

hominlee-wemade (approved):
LGTM

PR #186, opened 2026-05-08
fix: VerifyHeaders for mixed header batch at Croissant fork boundary

eomti-wm:
## Background

`VerifyHeaders` in each engine passes `headers[:i]` as in-memory parents when verifying each header in a batch. When a batch spans the Croissant fork boundary, the previous implementation fell into the mixed-batch path and called `VerifyHeader` individually for each header, providing no in-memory parents. As a result, `verifyCascadingFields` and `GetValidatorsForVerifying` fell back to `chain.GetHeader` for parent lookup. During snap sync, headers in the current batch are not yet committed to the chain DB, so every wbft header in a mixed batch failed with `ErrUnknownAncestor`.

## Solution

Split mixed batches at the Croissant boundary into legacy and wbft slices, and verify each slice with the appropriate engine as a batch. This restores the `headers[:i]` in-memory parent chain within each slice, eliminating chain DB fallback for headers in the current batch.

Since the wbft slice starts at the Croissant block, its first header's parent (the last legacy header) falls outside the slice and still requires a chain DB lookup. To address this, `VerifyHeadersWithParent` is added to the wbft backend to accept an explicit parent for `headers[0]`, used as the in-memory parent seed instead of falling back to the chain DB.

## Changes

`consensus/wemix/consensus.go`
- Split mixed batches at the Croissant boundary and verify each slice with the appropriate engine sequentially
- Pass the last legacy header as an explicit parent seed via `VerifyHeadersWithParent` for the wbft slice

`consensus/wbft/backend/engine.go`
- Add `VerifyHeadersWithParent(chain, headers, parent)` to accept an explicit parent for `headers[0]`
- Refactor `VerifyHeaders` and `VerifyHeadersWithParent` to share a common internal `verifyHeaders` implementation

### DIFF
--- a/consensus/wbft/backend/engine.go
+++ b/consensus/wbft/backend/engine.go
@@ -85,6 +85,21 @@ func (sb *Backend) verifyHeader(chain consensus.ChainHeaderReader, header *types
 // a results channel to retrieve the async verifications (the order is that of
 // the input slice).
 func (sb *Backend) VerifyHeaders(chain consensus.ChainHeaderReader, headers []*types.Header) (chan<- struct{}, <-chan error) {
+	return sb.verifyHeaders(chain, headers, nil)
+}
+
+// VerifyHeadersWithParent is like VerifyHeaders but accepts an explicit parent
+// header as the parent of headers[0]. Use this when the immediate parent of the
+// first header in the batch is not yet in the chain DB (e.g. snap sync batches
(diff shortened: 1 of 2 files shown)


hominlee-wemade (approved):
LGTM

colinkim (approved):
LGTM

0xmhha (approved):
LGTM

PR #192, opened 2026-06-08
fix(txpool): restore cumulative overdraft check for fee delegation txs

colinkim:
## Summary

Fee delegation tx (type 0x16) splits cost between sender (value) and feePayer (gas fee). The upstream cumulative overdraft check(introduced in ethereum/go-ethereum#27429) assumed a single payer via `totalcost`, causing false positives for valid fee delegation txs. As a workaround the check was globally disabled, leaving the pool exposed to two DoS vectors:

1. A single sender submitting sequential txs whose aggregate value exceeds balance — each passes per-tx validation individually but reverts on execution, polluting the mempool.
2. Multiple senders designating the same feePayer, collectively exceeding the feePayer's balance with no pool-level rejection.

## Changes

- **list.go**: add `totalvalue`, `totalgas`, `pendingGas` (pool-wide shared map); implement `addCost`/`subCost`. `pendingGas` aggregates gas obligations across all sender lists sharing the same feePayer, enabling cross-sender tracking.
- **legacypool.go**: inject `pendingGas` by reference into each pending list (pending-only). `ExistingExpenditure` returns total obligation across three roles: sender (`totalvalue`), own gas payer (`totalgas`), fee payer for others (`pendingGas`). `ExistingTx` added for correct replacement accounting.
- **validation.go**: replace disabled block with unified cumulative check via `need()`. Normal txs: combined balance check. Fee delegation txs: sender/feePayer checked independently.
- **blobpool_test.go**: restore upstream `core.ErrInsufficientFunds` expectations.
- **legacypool_test.go**: add `TestFeeDelegationCumulativeGas`.

## Test

- [x] `go test ./core/txpool/legacypool/ -run TestFeeDelegationCumulativeGas -v`
- [x] `go test ./core/txpool/legacypool/ ./core/txpool/blobpool/`
- [x] `go test -race ./core/txpool/legacypool/`

### DIFF
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -692,10 +692,16 @@ func (pool *LegacyPool) subPendingGas(payer common.Address, feeCost *uint256.Int
 	}
 }
 
-// validateTx checks whether a transaction is valid according to the consensus
-// rules and adheres to some heuristic limits of the local node (price and size).
-func (pool *LegacyPool) validateTx(tx *types.Transaction, local bool) error {
-	opts := &txpool.ValidationOptionsWithState{
+// stateValidationOptions builds the stateful validation options (nonce, balance
+// and cumulative overdraft checks) used by ValidateTransactionWithState. The same
+// option set is reused on submission (validateTx) and on promotion of queued
+// transactions, so the cumulative affordability check stays identical across both
+// paths. The callbacks read pool.pending / pool.pendingGas live, so when the
+// options are reused while promoting a batch, each freshly promoted transaction's
+// obligation is reflected in the next transaction's check.
+// Note, this method assumes the pool lock is held!
+func (pool *LegacyPool) stateValidationOptions() *txpool.ValidationOptionsWithState {
+	return &txpool.ValidationOptionsWithState{
 		State: pool.currentState,
 
 		FirstNonceGap:    nil, // Pool allows arbitrary arrival order, don't invalidate nonce gaps
@@ -731,7 +737,12 @@ func (pool *LegacyPool) validateTx(tx *types.Transaction, local bool) error {
 			return nil
 		},
 	}
-	if err := txpool.ValidateTransactionWithState(tx, pool.signer, opts); err != nil {
+}
+
+// validateTx checks whether a transaction is valid according to the consensus
+// rules and adheres to some heuristic limits of the local node (price and size).
+func (pool *LegacyPool) validateTx(tx *types.Transaction, local bool) error {
+	if err := txpool.ValidateTransactionWithState(tx, pool.signer, pool.stateValidationOptions()); err != nil {
 		return err
 	}
 	return pool.validateAuth(tx)
@@ -1613,8 +1624,12 @@ func (pool *LegacyPool) promoteExecutables(accounts []common.Address) []*types.T
 			pool.all.Remove(hash)
 		}
 		log.Trace("Removed old queued transactions", "count", len(forwards))
-		// Drop all transactions that are too costly (low balance or out of gas)
-		drops, _ := list.Filter(pool.chainconfig.IsApplepie(pool.currentHead.Load().Number), pool.currentState, pool.currentState.GetBalance(addr), gasLimit)
+		// Drop all transactions that are too costly (low balance or out of gas).
+		// enforceFeePayer is false here: a queued fee-delegated tx whose fee payer
+		// is only temporarily insolvent is left queued (and retried on a later
+		// pass) instead of being dropped. The promotion loop below still rejects it
+		// via ValidateTransactionWithState until the fee payer can afford it.
+		drops, _ := list.Filter(pool.chainconfig.IsApplepie(pool.currentHead.Load().Number), false, pool.currentState, pool.currentState.GetBalance(addr), gasLimit)
 		for _, tx := range drops {
 			hash := tx.Hash()
 			pool.all.Remove(hash)
@@ -1622,16 +1637,43 @@ func (pool *LegacyPool) promoteExecutables(accounts []common.Address) []*types.T
 		log.Trace("Removed unpayable queued transactions", "count", len(drops))
 		queuedNofundsMeter.Mark(int64(len(drops)))
 
-		// Gather all executable transactions and promote them
-		readies := list.Ready(pool.pendingNonces.get(addr))
+		// Peek the executable prefix without removing it, then admit only the
+		// transactions that pass. For fee-delegated transactions the shared fee
+		// payer may be over-committed once several senders' queues are promoted in
+		// the same pass, so run ValidateTransactionWithState (cumulative check)
+		// before each admit. On the first failure we stop: that transaction and
+		// every later one depend on it by nonce, so they stay queued (Peek-then-
+		// Admit) and are retried on a later pass once the fee payer can afford
+		// them. Plain transactions keep the existing per-tx behaviour.
+		readies := list.PeekReady(pool.pendingNonces.get(addr))
+		var (
+			admitted types.Transactions
+			fdOpts   *txpool.ValidationOptionsWithState // nil until the first FD tx; reused for subsequent ones
+		)
 		for _, tx := range readies {
-			hash := tx.Hash()
-			if pool.promoteTx(addr, hash, tx) {
+			if tx.Type() == types.FeeDelegateDynamicFeeTxType {
+				if fdOpts == nil {
+					fdOpts = pool.stateValidationOptions()
+				}
+				if err := txpool.ValidateTransactionWithState(tx, pool.signer, fdOpts); err != nil {
+					break // failing tx and its tail stay queued (nonce dependency)
+				}
+			}
+			if pool.promoteTx(addr, tx.Hash(), tx) {
 				promoted = append(promoted, tx)
+				admitted = append(admitted, tx)
+			} else {
+				// promoteTx already cleaned pool.all/priced; drop it from the
+				// queue and stop, since the tail depends on this nonce.
+				admitted = append(admitted, tx)
+				break
 			}
 		}
-		log.Trace("Promoted queued transactions", "count", len(promoted))
-		queuedGauge.Dec(int64(len(readies)))
+		// PeekReady did not remove anything; remove only the admitted prefix so the
+		// rejected tail remains in the queue.
+		list.RemoveReadyPrefix(admitted)
+		log.Trace("Promoted queued transactions", "count", len(admitted))
+		queuedGauge.Dec(int64(len(admitted)))
 
 		// Drop all transactions over the allowed limit
 		var caps types.Transactions
@@ -1815,7 +1857,9 @@ func (pool *LegacyPool) demoteUnexecutables() {
 			log.Trace("Removed old pending transaction", "hash", hash)
 		}
 		// Drop all transactions that are too costly (low balance or out of gas), and queue any invalids back for later
-		drops, invalids := list.Filter(pool.chainconfig.IsApplepie(pool.currentHead.Load().Number), pool.currentState, pool.currentState.GetBalance(addr), gasLimit)
+		// enforceFeePayer is true here: pending txs must stay payable, so a
+		// fee-delegated tx whose fee payer can no longer afford the gas is dropped.
+		drops, invalids := list.Filter(pool.chainconfig.IsApplepie(pool.currentHead.Load().Number), true, pool.currentState, pool.currentState.GetBalance(addr), gasLimit)
 		for _, tx := range drops {
 			hash := tx.Hash()
 			pool.all.Remove(hash)
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -670,7 +670,7 @@ func (pool *LegacyPool) addPendingGas(payer common.Address, feeCost *uint256.Int
 func (pool *LegacyPool) subPendingGas(payer common.Address, feeCost *uint256.Int) {
 	acc := pool.pendingGas[payer]
 	if acc == nil {
-		panic("pendingGas underflow")
+		panic("pendingGas: no entry for payer")
 	}
 	if _, underflow := acc.SubOverflow(acc, feeCost); underflow {
 		panic("pendingGas underflow")
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -1029,16 +1029,23 @@ func (pool *LegacyPool) journalTx(from common.Address, tx *types.Transaction) {
 	}
 }
 
+func newPendingList(add, sub func(common.Address, *uint256.Int)) *list {
+	if add == nil || sub == nil {
+		panic("legacypool: pending list requires both gas accounting callbacks")
+	}
+	l := newList(true)
+	l.addPendingGas = add
+	l.subPendingGas = sub
+	return l
+}
+
 // promoteTx adds a transaction to the pending (processable) list of transactions
 // and returns whether it was inserted or an older was better.
 // Note, this method assumes the pool lock is held!
 func (pool *LegacyPool) promoteTx(addr common.Address, hash common.Hash, tx *types.Transaction) bool {
 	// Try to insert the transaction into the pending queue
 	if pool.pending[addr] == nil {
-		list := newList(true)
-		list.addPendingGas = pool.addPendingGas
-		list.subPendingGas = pool.subPendingGas
-		pool.pending[addr] = list
+		pool.pending[addr] = newPendingList(pool.addPendingGas, pool.subPendingGas)
 	}
 	list := pool.pending[addr]
 
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -250,7 +250,7 @@ type LegacyPool struct {
 	reserver   txpool.Reserver                 // Address reserver to ensure exclusivity across subpools
 	pending    map[common.Address]*list        // All currently processable transactions
 	queue      map[common.Address]*list        // Queued but non-processable transactions
-	pendingGas map[common.Address]*uint256.Int // Cumulative pending gas fee indexed by gas payer (sender, or fee payer for delegation txs)
+	pendingGas map[common.Address]*uint256.Int // Cumulative pending gas fee indexed by fee payer for fee-delegated txs
 	beats      map[common.Address]time.Time    // Last heartbeat from each known account
 	all        *lookup                         // All transactions to allow lookups
 	priced     *pricedList                     // All transactions sorted by price
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -904,7 +904,11 @@ func (pool *LegacyPool) add(tx *types.Transaction, local bool) (replaced bool, e
 	// Try to replace an existing transaction in the pending pool
 	if list := pool.pending[from]; list != nil && list.Contains(tx.Nonce()) {
 		// Nonce already pending, check if required price bump is met
-		inserted, old := list.Add(tx, pool.config.PriceBump)
+		inserted, old, err := list.Add(tx, pool.config.PriceBump)
+		if err != nil {
+			pendingDiscardMeter.Mark(1)
+			return false, err
+		}
 		if !inserted {
 			pendingDiscardMeter.Mark(1)
 			return false, txpool.ErrReplaceUnderpriced
@@ -978,7 +982,11 @@ func (pool *LegacyPool) enqueueTx(hash common.Hash, tx *types.Transaction, local
 	if pool.queue[from] == nil {
 		pool.queue[from] = newList(false)
 	}
-	inserted, old := pool.queue[from].Add(tx, pool.config.PriceBump)
+	inserted, old, err := pool.queue[from].Add(tx, pool.config.PriceBump)
+	if err != nil {
+		queuedDiscardMeter.Mark(1)
+		return false, err
+	}
 	if !inserted {
 		// An older transaction was better, discard this
 		queuedDiscardMeter.Mark(1)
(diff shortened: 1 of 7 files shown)


egonspace (approved):
LGTM

core/txpool/legacypool/list.go
```diff
@@ -277,16 +277,36 @@ type list struct {
 	costcap   *uint256.Int // Price of the highest costing transaction (reset only if exceeds balance)
 	gascap    uint64       // Gas limit of the highest spending transaction (reset only if exceeds block limit)
 	totalcost *uint256.Int // Total cost of all transactions in the list
```
eomti-wm: Cost tracking has been split into `totalvalue` and `totalgas` with this commit,
but `totalcost` is still being accumulated and decremented. Since the split
makes `totalcost` redundant — it is written in `Add()` and `subTotalCost()` but
never read anywhere — removing it would be the right direction.

core/txpool/legacypool/list.go
```diff
@@ -471,6 +492,60 @@ func (l *list) subTotalCost(txs []*types.Transaction) {
...
 		}
+		l.subCost(tx)
+	}
+}
+
+// addCost attributes a transaction's value and gas obligations to the right
+// accounts: the value is always owed by the sender (this list's owner), while
+// the gas is owed by the sender for normal transactions or by the fee payer for
+// fee-delegated transactions (tracked pool-wide via pendingGas).
+func (l *list) addCost(tx *types.Transaction) {
```
eomti-wm: The nil guard for `pendingGas` is placed inside the fee-delegated branch. If
`pendingGas` is nil, `totalvalue` is still updated while the gas obligation is
silently skipped, causing `ExistingExpenditure` to underreport the obligation
with only value accumulated and gas ignored.

This does not occur under the current flow, but some improvements seem
necessary for the guard to work correctly — such as guard position, prevention
of silent failure (error logging, error return, success/failure signalling,
etc.). 

Same applies to `subCost`.

core/txpool/legacypool/legacypool.go
```diff
@@ -247,12 +247,13 @@ type LegacyPool struct {
...
-	reserver txpool.Reserver              // Address reserver to ensure exclusivity across subpools
-	pending  map[common.Address]*list     // All currently processable transactions
-	queue    map[common.Address]*list     // Queued but non-processable transactions
-	beats    map[common.Address]time.Time // Last heartbeat from each known account
-	all      *lookup                      // All transactions to allow lookups
-	priced   *pricedList                  // All transactions sorted by price
+	reserver   txpool.Reserver                 // Address reserver to ensure exclusivity across subpools
+	pending    map[common.Address]*list        // All currently processable transactions
+	queue      map[common.Address]*list        // Queued but non-processable transactions
+	pendingGas map[common.Address]*uint256.Int // Cumulative pending gas fee indexed by gas payer (sender, or fee payer for delegation txs)
```
eomti-wm: The comment says "sender, or fee payer for delegation txs", but normal
transaction gas is never written to this map — it goes to `list.totalgas`
instead. The comment is inaccurate and should be corrected.

core/txpool/legacypool/legacypool.go
```diff
@@ -988,7 +1006,12 @@ func (pool *LegacyPool) journalTx(from common.Address, tx *types.Transaction) {
 func (pool *LegacyPool) promoteTx(addr common.Address, hash common.Hash, tx *types.Transaction) bool {
 	// Try to insert the transaction into the pending queue
 	if pool.pending[addr] == nil {
-		pool.pending[addr] = newList(true)
+		list := newList(true)
+		// Share the pool-wide fee-payer gas index by reference so this sender's
+		// fee-delegated txs contribute to their fee payer's cumulative obligation.
+		// Only pending lists are wired up (the overdraft check is pending-only).
+		list.pendingGas = pool.pendingGas
```
eomti-wm: `list` holds a direct pointer to `pool.pendingGas` and mutates it in place,
giving a child component write access to the parent's internal state. This
couples `list` to the pool's data structure and breaks encapsulation.

Passing a callback function to `list` instead, so that `pendingGas` ownership
and updates remain solely within the pool, would be worth considering.

core/txpool/validation.go
```diff
@@ -267,34 +279,98 @@ func ValidateTransactionWithState(tx *types.Transaction, signer types.Signer, op
...
+			// Pools without fee-delegation support (e.g. the blob pool): the replaced
+			// transaction is fully sender-paid, so its whole cost belongs to the sender.
+			hasPrev = true
+			prevGasCost = prev // prev == old.Cost(); oldGasPayer == from
+		}
+
+		// need computes an account's total pooled obligation (value owed as a
+		// sender plus gas owed as a gas payer), augmented by the incoming
+		// transaction's deltas and discounted by any transaction being replaced.
+		need := func(addr common.Address, addValue, addGas *big.Int) *big.Int {
```
eomti-wm: `addValue` and `addGas` do not make it clear that they refer to the incoming
transaction's own value and gas. The function already uses `prevValue` and
`prevGasCost` for the replaced transaction, so naming the counterparts
`curValue` and `curGasCost` would make the prev/cur relationship explicit and
easier to follow.

core/txpool/legacypool/list.go
```diff
@@ -471,6 +492,60 @@ func (l *list) subTotalCost(txs []*types.Transaction) {
...
+	l.totalvalue.Add(l.totalvalue, uint256.MustFromBig(tx.Value()))
+
+	feeCost := uint256.MustFromBig(tx.FeeCost())
+	if tx.Type() == types.FeeDelegateDynamicFeeTxType && tx.FeePayer() != nil {
```
hmlee-wm: If tx.Type() == FeeDelegateDynamicFeeTxType but tx.FeePayer() == nil, this function falls through to totalgas and attributes the gas cost to the sender rather than the fee payer.

This should be unreachable through the normal txpool path because ValidateTransaction() rejects a nil fee payer for fee-delegated txs. 

Since list.Add() does not enforce this invariant, it may be clearer to check FeePayer() == nil explicitly and fail fast rather than silently misattributing the gas cost.

core/txpool/legacypool/list.go
```diff
@@ -471,6 +492,60 @@ func (l *list) subTotalCost(txs []*types.Transaction) {
...
+}
+
+// subCost reverses addCost for a removed transaction.
+func (l *list) subCost(tx *types.Transaction) {
+	if _, underflow := l.totalvalue.SubOverflow(l.totalvalue, uint256.MustFromBig(tx.Value())); underflow {
+		panic("totalvalue underflow")
+	}
+
+	feeCost := uint256.MustFromBig(tx.FeeCost())
+	if tx.Type() == types.FeeDelegateDynamicFeeTxType && tx.FeePayer() != nil {
```
hmlee-wm: Mirrors addCost: nil FeePayer() falls through to totalgas.

core/txpool/legacypool/list.go
```diff
@@ -471,6 +492,60 @@ func (l *list) subTotalCost(txs []*types.Transaction) {
...
+// addCost attributes a transaction's value and gas obligations to the right
+// accounts: the value is always owed by the sender (this list's owner), while
+// the gas is owed by the sender for normal transactions or by the fee payer for
+// fee-delegated transactions (tracked pool-wide via pendingGas).
+func (l *list) addCost(tx *types.Transaction) {
+	l.totalvalue.Add(l.totalvalue, uint256.MustFromBig(tx.Value()))
+
+	feeCost := uint256.MustFromBig(tx.FeeCost())
+	if tx.Type() == types.FeeDelegateDynamicFeeTxType && tx.FeePayer() != nil {
+		if l.pendingGas == nil {
```
hmlee-wm: totalvalue has already been updated above. Returning here would leave the fee-delegated gas untracked, so it may be better to fail fast instead of silently returning.

core/txpool/legacypool/list.go
```diff
@@ -471,6 +492,60 @@ func (l *list) subTotalCost(txs []*types.Transaction) {
...
+// subCost reverses addCost for a removed transaction.
+func (l *list) subCost(tx *types.Transaction) {
+	if _, underflow := l.totalvalue.SubOverflow(l.totalvalue, uint256.MustFromBig(tx.Value())); underflow {
+		panic("totalvalue underflow")
+	}
+
+	feeCost := uint256.MustFromBig(tx.FeeCost())
+	if tx.Type() == types.FeeDelegateDynamicFeeTxType && tx.FeePayer() != nil {
+		if l.pendingGas == nil {
+			return
```
hmlee-wm: totalvalue has already been updated above. Returning here would leave the fee-delegated gas untracked, so it may be better to fail fast instead of silently returning.

core/txpool/validation.go
```diff
@@ -267,34 +279,98 @@ func ValidateTransactionWithState(tx *types.Transaction, signer types.Signer, op
...
+		}
+		txValue := tx.Value()
+		txGasCost := tx.FeeCost() // tx.Cost() - tx.Value()
+
+		// If a transaction with the same sender and nonce is already pooled, the
+		// incoming one replaces it, so its contribution must be discounted from
+		// whichever account currently carries it (value -> sender, gas -> its own
+		// gas payer, which may differ from the new transaction's gas payer).
+		var (
+			hasPrev      bool
```
hmlee-wm: hasPrev indicates that the incoming tx replaces an existing pending tx with the same sender/nonce. Renaming it to isReplacement would make the subtraction of prevValue and prevGasCost in need() easier to follow.

core/txpool/validation.go
```diff
@@ -267,34 +279,98 @@ func ValidateTransactionWithState(tx *types.Transaction, signer types.Signer, op
...
+			n.Add(n, addGas)
+			if hasPrev {
+				if addr == from {
+					n.Sub(n, prevValue)
+				}
+				if addr == prevGasPayer {
+					n.Sub(n, prevGasCost)
+				}
+				if n.Sign() < 0 {
+					n.SetInt64(0) // guard against accounting drift
```
hmlee-wm: If accounting is consistent, need() should not produce a negative value here.

A negative value means the replacement discount is larger than the account's recorded obligation, which indicates an accounting inconsistency — ExistingExpenditure undercounts what the replaced transaction actually cost.

Clamping it to zero hides that invariant violation and allows validation to continue with a corrupted accounting state.

Could we fail fast or at least return an explicit error instead of silently normalizing this case?

hominlee-wemade:
This PR changes txpool accounting on the add/remove path, so benchmarks would help confirm there's no regression.

core/txpool/validation.go
colinkim: fixed in f573a97

core/txpool/validation.go
colinkim: fixed in 1a69136

core/txpool/validation.go
colinkim: fixed in f573a97

core/txpool/legacypool/list.go
colinkim: Precondition — ValidateTransaction guarantees FeePayer != nil forfee-delegated txs before list.Add is reached, so the fall-throughpath is unreachable in practice. Kept the outer && FeePayer != nilas a belt-and-suspenders guard;

core/txpool/legacypool/list.go
colinkim: Precondition — ValidateTransaction guarantees FeePayer != nil forfee-delegated txs before list.Add is reached, so the fall-throughpath is unreachable in practice. Kept the outer && FeePayer != nilas a belt-and-suspenders guard;

core/txpool/legacypool/list.go
colinkim: fixed in 320eab6

core/txpool/legacypool/legacypool.go
colinkim: fixed in d8d7623

core/txpool/legacypool/list.go
```diff
@@ -463,14 +482,63 @@ func (l *list) LastElement() *types.Transaction {
...
+		}
+		payer := *tx.FeePayer()
+		acc := l.pendingGas[payer]
+		if acc == nil {
+			panic("pendingGas underflow")
+		}
```
eomti-wm: If `acc` is nil, the panic message says "pendingGas underflow", but an underflow
has not occurred — the map entry for the fee payer simply does not exist. Without
this guard, the next line (`acc.SubOverflow(...)`) would produce a
`nil pointer dereference` error, not an underflow.
The message should be corrected to reflect the actual condition.

core/txpool/legacypool/list.go
hmlee-wm: Stale comments remain from the old totalcost-based accounting:
- Add new tx cost to totalcost
- Reset total cost

core/txpool/legacypool/list.go
colinkim: This is intentional. totalvalue is updated before the return on purpose: the matching subCost (called from subTotalCost) subtracts totalvalue unconditionally, so addCost must add it the same way — returning before the add would make the subtraction underflow and panic. The fee-delegated gas being skipped here is also intentional: queue lists are excluded from the pending-only overdraft accounting, and the gas is recorded once the tx is promoted to a pending list.

core/txpool/legacypool/list.go
colinkim: This is intentional. totalvalue is updated before the return on purpose: the matching subCost (called from subTotalCost) subtracts totalvalue unconditionally, so addCost must add it the same way — returning before the add would make the subtraction underflow and panic. The fee-delegated gas being skipped here is also intentional: queue lists are excluded from the pending-only overdraft accounting, and the gas is recorded once the tx is promoted to a pending list.

core/txpool/legacypool/list.go
colinkim: This is intentional. totalvalue is updated before the return on purpose: the matching subCost (called from subTotalCost) subtracts totalvalue unconditionally, so addCost must add it the same way — returning before the add would make the subtraction underflow and panic. The fee-delegated gas being skipped here is also intentional: queue lists are excluded from the pending-only overdraft accounting, and the gas is recorded once the tx is promoted to a pending list.

core/txpool/legacypool/legacypool.go
colinkim: fixed in 8f473c3

core/txpool/legacypool/list.go
colinkim: fixed in 8f473c3

core/txpool/legacypool/list.go
colinkim: Updated in 8f473c3

core/txpool/legacypool/list.go
colinkim: Updated in 8f473c3 for fail fast

core/txpool/legacypool/list.go
colinkim: Updated in 8f473c3 for fail fast

core/txpool/legacypool/list.go
eomti-wm: The logic was moved from `list.go` into `subPendingGas`, but the panic message is still incorrect.

core/txpool/legacypool/list.go
colinkim: fixed in 8f473c3

core/txpool/legacypool/list.go
colinkim: fixed in 8f473c3

core/txpool/legacypool/list.go
colinkim: fixed in bdfe73dc

core/txpool/legacypool/list.go
colinkim: fixed in f348811

core/txpool/legacypool/list.go
colinkim: fixed in 254e861

core/txpool/legacypool/list.go
colinkim: fixed in 254e861

core/txpool/legacypool/list.go
colinkim: d3ac377a8 has taken the explanation from the previous 8f473c3 (symmetry of totalvalue, exclusion of queue-list) and expanded it. The key difference is that the responsibility for "fail fast" has been moved up to the list.Add entry instead of placing it in the in-branch guard.

1. Guard position (Moved up)

The list.Add entry rejects any state mutation (subCosts/addCost/Put) before any other. Therefore, fee-delegated transactions with nil FeePayer are rejected before totalvalue is added, preventing the "totalvalue already updated → gas untracked" state—which was a concern in the old thread—from forming.

2. Signalling

Extended the list.Add signature to (bool, *Transaction, error). - add(), enqueueTx(): Surface ErrInvalidFeePayer as is (previously likely mislabeled as ErrReplaceUnderpriced)

- promoteTx(): Since it is a boolean return function, error propagation is not possible → Explicitly logging via log.Error

3. in-branch guard for addCost/subCost

The existing in-branch nil guard has been maintained. This is because the totalvalue symmetry and queue-list policy explained in the old thread remain valid. However, since the ErrInvalidFeePayer and list.Add entry guards in validation.go:123, both of which block the same case, preemptively reject the same case, the in-branch nil branch is now a defensive backstop that is likely unreachable. In other words, the concern regarding a silent return changes to "a silent return from an unreachable path." If you still believe that explicit fail-fast at the in-branch stage is necessary, I think it would be better to change the log.Error + return of addCost/subCost to panic.

colinkim: fixed in 254e861

core/txpool/legacypool/list.go
colinkim: fixed in 254e861

core/txpool/legacypool/list.go
colinkim: fixed in 254e861

core/txpool/legacypool/legacypool.go
colinkim: updated in ab21a91

eomti-wm:
The comment says "re-checked on promotion", but the re-check is incomplete for fee-delegated transactions.

On submission, validateTx checks cumulative affordability using pendingGas[feePayer]. On promotion, list.Filter() only checks whether the fee payer can afford the single tx in isolation. It does not include the fee payer's existing pending obligation or other ready txs promoted in the same pass.

So queued txs can pass promotion individually even when the fee payer's total obligation after promotion exceeds its balance.

This is different from Ethereum's single-payer model. There, the queue is keyed by sender and the payer is also the sender, so sender-local filtering is enough to reason about that sender's queued txs. With fee delegation, the queue is still keyed by sender, but gas is paid by feePayer, and the same feePayer can be shared across many sender queues.

The promotion path should apply the same cumulative check as validateTx, or simulate fee payer obligations incrementally before admitting ready txs.

core/txpool/legacypool/list.go
```diff
@@ -463,14 +486,65 @@ func (l *list) LastElement() *types.Transaction {
...
+// are always set as a pair — if only one were nil, addCost/subCost would diverge
+// and the accounting (totalvalue/pendingGas) would become wrong, so a half-wired
+// state is treated as a fatal wiring bug and panics. Used as the top-level guard
+// in addCost/subCost so both stay symmetric.
+func (l *list) tracksExpenditure() bool {
+	if (l.addPendingGas == nil) != (l.subPendingGas == nil) {
+		panic("inconsistent pendingGas callbacks would corrupt gas accounting")
+	}
+	return l.addPendingGas != nil
+}
```
hmlee-wm, line 523: nit: It might be clearer to split the callback nil checks into explicit variables.

The current logic is correct, but `(l.addPendingGas == nil) != (l.subPendingGas == nil)` requires reading it as an XOR comparison of nil states.

Using `hasAddPendingGas` and `hasSubPendingGas` makes the invariant clearer: both callbacks must be configured as a pair.

```suggestion

func (l *list) tracksExpenditure() bool {
	hasAddPendingGas := l.addPendingGas != nil
	hasSubPendingGas := l.subPendingGas != nil
	if hasAddPendingGas != hasSubPendingGas {
		panic("inconsistent pendingGas callbacks would corrupt gas accounting")
	}
	return hasAddPendingGas
}
```

core/txpool/legacypool/list.go
eomti-wm: On second thought, rather than splitting the cost tracking into `totalvalue` and `totalgas` (and removing `totalcost` as I suggested above), it would be much simpler and more memory-efficient to revert the split entirely and reuse the existing `totalcost` field.

We can redefine `totalcost` as "the cumulative cost the sender is directly responsible for":
- Normal transactions: `Value + Gas` (i.e. `tx.Cost()`)
- Fee-delegated transactions: `Value` only (gas is routed to the fee payer's `pendingGas` entry)

This keeps a single `*uint256.Int` field per `list` instead of two, and `ExistingExpenditure` can return `list.totalcost + pendingGas[addr]` directly.

colinkim:
> The comment says "re-checked on promotion", but the re-check is incomplete for fee-delegated transactions.
> 
> On submission, validateTx checks cumulative affordability using pendingGas[feePayer]. On promotion, list.Filter() only checks whether the fee payer can afford the single tx in isolation. It does not include the fee payer's existing pending obligation or other ready txs promoted in the same pass.
> 
> So queued txs can pass promotion individually even when the fee payer's total obligation after promotion exceeds its balance.
> 
> This is different from Ethereum's single-payer model. There, the queue is keyed by sender and the payer is also the sender, so sender-local filtering is enough to reason about that sender's queued txs. With fee delegation, the queue is still keyed by sender, but gas is paid by feePayer, and the same feePayer can be shared across many sender queues.
> 
> The promotion path should apply the same cumulative check as validateTx, or simulate fee payer obligations incrementally before admitting ready txs.

fixed in e005ab6
updated in 3a4088c

core/txpool/legacypool/list.go
colinkim: fixed in 5033f1f

colinkim:
Replaced with new PR

hominlee-wemade:
During the review process, the implementation evolved significantly from the original design and went through multiple iterations.

To make the final implementation easier to review with a cleaner history, this PR will be closed, and a new follow-up PR containing the finalized implementation will be created.

Please continue the review in the new PR once it is available.

PR #195, opened 2026-06-17
fix: harden istanbul_status RPC against resource exhaustion and data integrity issues

eomti-wm:
## Overview
This PR hardens the `istanbul_status` RPC API in the WBFT consensus module by introducing block range limits and epoch-based validator set caching, resolving several reliability issues and resource exhaustion risks.

## Problem
- The lack of an upper limit on the requested block range allowed unbounded database scans, exposing the node to CPU and DB I/O resource exhaustion.
- Casting negative block number constants (e.g. `LatestBlockNumber`) directly to `uint64` caused underflow (wrapping), leading to incorrect parsing.
- Failures during block analysis silently fallback to round 0, polluting distribution statistics and causing partial counter updates.
- Inactive validators at epoch boundaries were omitted from the final response, and redundant database header queries occurred on every block.

## Solution
- Introduced a maximum request limit constant (`maxStatusBlockRange`) in `calculateBlockRange` to restrict scan ranges, and correctly mapped `LatestBlockNumber` while rejecting other unsupported negative constants.
- Refactored `analyzeBlock` to return errors explicitly and deferred counter increments until all data fetches succeed.
- Implemented local caching for validator sets refreshed only on epoch boundaries, initializing zero baseline entries dynamically for all active/inactive validators.

## Changes
- Add range validation and explicit mapping for `rpc.BlockNumber` constants in `calculateBlockRange`
- Redesign `analyzeBlock` to return `(uint64, error)` and defer counter updates to prevent partial state corruption
- Implement local caching of validator sets and zero-baseline initialization in the `Status` loop
- Remove the ambiguous `TotalRounds` field from `RoundStats` struct

### DIFF
--- a/consensus/wbft/backend/api.go
+++ b/consensus/wbft/backend/api.go
@@ -69,7 +69,6 @@ type BlockRange struct {
 // RoundStats contains round distribution statistics
 type RoundStats struct {
 	RoundDistribution map[uint64]uint64 `json:"roundDistribution"` // Map of round number to occurrence count
-	TotalRounds       uint64            `json:"totalRounds"`       // Total number of rounds processed
 }
 
 // Status contains validator activity statistics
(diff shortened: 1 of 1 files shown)


hominlee-wemade (approved):
LGTM

0xmhha (approved):
LGTM

colinkim (approved):
In `analyzeBlock`, the zero-baseline initialization for `PrevPrepared` /
`PrevCommitted` can be skipped for newly added validators after an epoch
transition.

For example, assume the epoch block is `100`.

Before the epoch transition, validators are:

```text
A, B, C
```
After the epoch transition, validators are:
```text
B, C, D
```
When block 101 is analyzed, the cache is refreshed because
cachedCurVals == nil.
At this point:
```go
curValidators  = B, C, D // new epoch validators
prevValidators = A, B, C // previous epoch validators
```
The code then initializes zero-baseline entries like this:
```go
// Initialize zero baseline here
for _, addr := range *cachedCurVals {
	initZero(addr, activity.Prepared, activity.Committed, activity.Total, authorCounts)
}
for _, addr := range *cachedPrevVals {
	initZero(addr, activity.PrevPrepared, activity.PrevCommitted, activity.Total, authorCounts)
}
```
So D is included as a new epoch validator:
```go
Prepared[D] = 0
Committed[D] = 0
```
But D is not part of the previous epoch validator set, so at this point these
entries are not created:
```go
PrevPrepared[D] = 0
PrevCommitted[D] = 0
```
After `analyzeBlock` finishes processing block `101`, the code runs:
```go
*cachedPrevVals = *cachedCurVals
```
This is correct because block 102's prev seals refer to parent block 101,
which was produced by the new epoch validators. From block 102, the prev
validator set should also be B, C, D.
However, when `cachedPrevVals` changes from `A, B, C` to `B, C, D`, the code
does not insert the newly added validator `D` into the `PrevPrepared` /
`PrevCommitted` maps with a zero value.
As a result, from block 102 onward:
- if D signs a prev seal, target[addr]++ creates the key and the count is
recorded correctly 
- if D does not sign, D remains absent from PrevPrepared /
PrevCommitted instead of being included with value 0 

So the count is correct when `D` signs. But if `D` is inactive and does not sign,`D` can be missing from the JSON response instead of being included with value`0`.

colinkim (approved):
LGTM

hominlee-wemade (approved):
LGTM

consensus/wbft/backend/api.go
```diff
@@ -167,185 +166,190 @@ func (api *API) GetValidatorsAtHash(hash common.Hash) ([]common.Address, error)
...
+		}
+		for _, addr := range *cachedCurVals {
+			initZero(addr, activity.Prepared, activity.Committed, activity.Total, authorCounts)
+		}
+		for _, addr := range *cachedPrevVals {
+			initZero(addr, activity.PrevPrepared, activity.PrevCommitted, activity.Total, authorCounts)
+		}
+		// curVals will become prevVals on the next block — pre-initialize PrevPrepared/PrevCommitted
+		// unless this is the last block (curVals will never be used as prevVals in the range).
+		if !isLastBlock {
```
0xmhha: I'd like to suggest a simplification regarding the `isLastBlock` parameter added to `analyzeBlock`.

Per its own doc comment, this function "analyzes a single block." By taking `isLastBlock`, however, it now has to be aware of the caller's loop boundary (`n == end`). Requiring a single-block analyzer to know the boundary of an external iteration is, in terms of coupling, a leaky abstraction.

If we work out when the guard actually affects the output, it only applies when `*cachedCurVals == nil` (i.e., the start block or the block right after an epoch transition) *and* it is simultaneously the last block — that is, the `end == epochBlock+1` case only. For a start block in the middle of an epoch, `prevVals == curVals`, so there is no difference. Moreover, in that case the value being added is always `0`, so it has no effect on the actual count totals such as Total/Prepared/Committed. It is also a far more forgiving failure mode than omission (the bug this commit fixed) — at worst it exposes one extra `0` entry.

So I propose folding the `curVals` → `PrevPrepared/PrevCommitted` initialization into the `cachedCurVals` loop at lines 293–295 and performing it unconditionally, then removing the `isLastBlock` parameter. The only behavioral difference is that, when `end == epochBlock+1`, the new-epoch validators are exposed in the prev maps as `0`; this has no effect on the counts, and it lowers the coupling of the function.

eomti-wm: fix : 86e176f3eaf2592a55f8a5350516b291f2479aa3

PR #196, opened 2026-06-18
fix: harden WBFT backlog against future-message flooding

eomti-wm:
## Overview

Prevent memory and CPU exhaustion DoS attacks by validating and restricting future-sequence messages in the WBFT backlog.

## Problem

Future-sequence messages (sequence = current + 1) were stored in the validator's backlog without round threshold validation nor per-validator queue size limits.

Additionally, the existing knownMessages LRU deduplicates by full payload hash, so a malicious validator can bypass it by varying fields such as Digest or Justification while keeping the same (sequence, round) coordinates, generating unlimited distinct messages for the same logical slot.

A malicious validator could spam messages to cause memory and CPU exhaustion, potentially causing liveness degradation.

## Solution

Enforce round threshold validation on future-sequence messages, cap the maximum backlog size per validator, and introduce a key-based deduplication map. Keep backlogKeys in sync by removing keys when messages are dropped or when a validator is removed.

## Changes

- Apply roundThreshold validation to future-sequence messages in isTooFarFutureMessage.
- Implement maxBacklogSizePerValidator cap to drop overflow messages in addToBacklog.
- Introduce backlogKey and backlogKeys map to prevent duplicate messages for the same slot in addToBacklog.
- Keep backlogKeys in sync by removing keys on message drop and purging all backlog metadata upon validator removal in processBacklog.


consensus/wbft/core/backlog.go
```diff
@@ -201,7 +225,24 @@ func (c *Core) addToBacklog(msg wbfmessage.WBFTMessage) {
...
+	// Drop the message if the same (code, sequence, round) slot is already queued for this validator.
+	if keys, ok := c.backlogKeys[src]; ok {
+		if _, exists := keys[pkey]; exists {
+			logger.Trace("WBFT: duplicate backlog message, dropping", "src", src)
+			return
+		}
+	} else {
+		c.backlogKeys[src] = make(map[backlogKey]struct{})
+	}
+	c.backlogKeys[src][pkey] = struct{}{}
```
hmlee-wm: Currently, when the backlog is already full, duplicate messages are rejected by the size check before reaching the duplicate check, resulting in backlog is full being logged instead of duplicate backlog message.

Performing the duplicate check first would make the rejection reason more accurate. In addition, delaying c.backlogKeys[src] initialization until after all validation checks would avoid unnecessary map allocation when the message is rejected.

```suggestion
view := msg.View()
pkey := backlogKey{msg.Code(), view.Sequence.Uint64(), view.Round.Uint64()}
// Drop the message if the same (code, sequence, round) slot is already queued for this validator.
if keys, ok := c.backlogKeys[src]; ok {
    if _, exists := keys[pkey]; exists {
        logger.Trace("WBFT: duplicate backlog message, dropping", "src", src)
        return
    }
}

// Reject messages from a validator whose backlog exceeds the size limit.
if backlog.Size() >= maxBacklogSizePerValidator {
    logger.Warn("WBFT: backlog is full, dropping message", "src", src, "size", backlog.Size())
    return
}

if c.backlogKeys[src] == nil {
    c.backlogKeys[src] = make(map[backlogKey]struct{})
}
c.backlogKeys[src][pkey] = struct{}{}
```

eomti-wm: fix : 9a2bc7d2825ebb5d0c4ca3f16de80a7ca0328029

consensus/wbft/core/backlog.go
```diff
@@ -44,8 +44,26 @@ var (
...
+	sequence uint64
+	round    uint64
+}
+
+// assertComparable enforces at compile time that T can be used as a map key.
+func assertComparable[T comparable]() {}
+
+// compile-time check: all fields of backlogKey must remain comparable for use as a map key.
+var _ = assertComparable[backlogKey]
+
```
hmlee-wm: The requirement that backlogKey remain comparable is already enforced by the compiler wherever it is used as a map key (e.g. map[backlogKey]struct{}). As a result, assertComparable effectively performs the same check again without providing any additional safety guarantees.

This pattern is only useful when there is an explicit need to enforce comparability at the type definition level — for example, when the type is intended to remain comparable before any map usage exists, or when other packages depend on that constraint.

In this case, backlogKey consists entirely of uint64 fields, making its comparability self-evident, and its map usage already exists within the same file, so those conditions do not apply here.

I’d suggest removing it to reduce unnecessary code.

```suggestion
```

eomti-wm: fix : 33e0335e6f75ec2006f52a360ee839d68e97b422

colinkim (approved):
LGTM

consensus/wbft/core/backlog.go
```diff
@@ -74,6 +86,15 @@ func (c *Core) isTooFarFutureMessage(view *wbft.View) bool {
 			)
 			return true
 		}
+		// Reject future sequence messages with an excessively high round number.
+		if view.Round.Cmp(big.NewInt(roundThreshold)) >= 0 {
```
0xmhha, line 91: **The future-seq round threshold is one round stricter than the same-seq check — could you confirm whether this is an intentional tightening?**

I converted both paths onto a common "diff basis" and compared them. Because the next sequence starts at round 0, the future-seq baseline is 0; therefore `view.Round` itself is precisely the diff.

| Path | Reject condition | Normalized diff | Allowed rounds |
|------|------------------|-----------------|----------------|
| same-seq | `view.Round - curr.Round > 10` | `diff > 10` | (curr=0) 0–10 |
| future-seq | `view.Round >= 10` | `diff > 9` | 0–9 |

Viewed on the same diff basis, same-seq is allowed up to 10 while future-seq is allowed up to 9, producing a one-round difference. This is likewise baked into the test (`TestIsTooFarFutureMessageRoundThreshold`): when round=10, future-seq is rejected (true) while same-seq is allowed (false).

Since the commit `3359ffe12` ("align future-sequence round limit with current-sequence check") message reads as an intent to align the two checks, I'd like to confirm which behavior is intended:

- **If alignment is the goal** (allow future-seq up to diff 10 as well) → change `>= roundThreshold` to `> roundThreshold`
- **If it is an intentional tightening** (round ≥ 10 on a next block that hasn't even started is abnormal, so allow only up to 9) → keep the code and state explicitly in a comment that "future-seq has baseline 0, so it is one round stricter than same-seq"

If the design intent is the latter, augmenting the comment without any code change should be sufficient.

eomti-wm: fix: b9d13a5da43426765cb21d1277c7d7bc260ce33e

The current implementation is indeed the intentional behavior. To clear up any potential ambiguity from the previous commit message regarding the boundary check numbers, the comment has been updated to explicitly state the design intent:

* **Same-sequence:** Since the current round is already underway, up to `roundThreshold` future rounds starting from `curr+1` through `curr+roundThreshold` are allowed. This is implemented via the diff-based check (`> roundThreshold`).
* **Future-sequence:** Since the future sequence starts fresh from round `0` and all its rounds are considered future rounds, allowing `roundThreshold` future rounds means allowing rounds `0` to `roundThreshold-1`. This is implemented via the absolute check (`>= roundThreshold`).

Both checks consistently allow exactly `roundThreshold` future rounds under their respective contexts.

0xmhha (approved):
LGTM

hominlee-wemade (approved):
LGTM

PR #197, opened 2026-06-21
fix(txpool): restore cumulative affordability enforcement with fee-delegation accounting

colinkim:
## Summary

Fee-delegated transactions (type `0x16`) split their financial obligations:

- the sender pays `tx.Value()`
- the fee payer pays `tx.FeeCost()`

The upstream cumulative overdraft check introduced by ethereum/go-ethereum#27429 assumed that the sender pays the entire transactioncost. Because that assumption caused false rejections for fee-delegated transactions, the check had been disabled globally.

Disabling it allowed individually affordable but collectively unfundable transactions to occupy txpool capacity:

1. A sender could submit sequential transactions whose cumulative cost exceeded its balance.
2. Multiple senders could charge gas to the same fee payer until the aggregate obligation exceeded the fee payer's balance.
3. Queued transactions could bypass the pending-only accounting and become collectively unaffordable when promoted together.

This change restores cumulative overdraft protection while accounting for sender and fee-payer obligations separately.

### Cumulative accounting

- Keep `list.totalcost` for sender-side pending obligations:
  - normal transaction: `tx.Cost()`
  - fee-delegated transaction: `tx.Value()`
- Add pool-wide `pendingGas`, indexed by fee payer, to aggregate delegated gas obligations across sender lists.
- Make pending lists update `pendingGas` through paired add/remove callbacks. Queued transactions remain outside pending expenditure accounting.
- Ignore zero gas obligations symmetrically and detect accounting underflows as internal invariant failures.

### Stateful validation

- Restore cumulative overdraft validation for normal transactions.
- Check fee-delegated sender value and fee-payer gas against their respective balances.
- Add `ExistingTx` so replacement validation can subtract the previous transaction's value and gas from the correct accounts, including fee-payer changes.
- Reuse the same state validation options for submission and queued promotion.

### Promotion and demotion

- Revalidate each ready fee-delegated transaction before promoting it from queue to pending.
- Account for transactions promoted earlier in the same pass when checking a shared fee payer.
- On the first promotion failure, drop the rejected transaction and its dependent nonce tail.
- Disable the list filter short circuit while fee-delegated transactions are present, ensuring a transaction is removed from pending if its fee payer later becomes unable to cover the gas obligation.

### Tests

- Restore blob-pool cumulative insufficient-funds expectations.
- Add coverage for:
  - cumulative delegated gas across multiple senders
  - cumulative validation during queued promotion
  - rejected nonce-tail removal
  - zero-value `pendingGas` accounting
  - fee-delegated list-counter consistency
  - filter behavior after fee-payer insolvency

## Test

- [x] `go test ./core/txpool/legacypool ./core/txpool/blobpool`
- [x] `go test -race ./core/txpool/legacypool`
- [x] `git diff --check origin/dev...HEAD`

## Related PR

This PR supersedes #192. See #192 for the original discussion and review history.

### DIFF
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -687,10 +687,13 @@ func (pool *LegacyPool) subPendingGas(payer common.Address, feeCost *uint256.Int
 	}
 }
 
-// validateTx checks whether a transaction is valid according to the consensus
-// rules and adheres to some heuristic limits of the local node (price and size).
-func (pool *LegacyPool) validateTx(tx *types.Transaction, local bool) error {
-	opts := &txpool.ValidationOptionsWithState{
+// stateValidationOptions builds the stateful validation options used by
+// ValidateTransactionWithState. Both submission and queued promotion use this
+// builder so nonce, balance, and cumulative overdraft checks stay identical.
+// The expenditure callbacks read pending state at call time, so promotion
+// reflects txs admitted earlier in the same batch. The pool lock must be held.
+func (pool *LegacyPool) stateValidationOptions() *txpool.ValidationOptionsWithState {
+	return &txpool.ValidationOptionsWithState{
 		State: pool.currentState,
 
 		FirstNonceGap:    nil, // Pool allows arbitrary arrival order, don't invalidate nonce gaps
@@ -726,7 +729,12 @@ func (pool *LegacyPool) validateTx(tx *types.Transaction, local bool) error {
 			return nil
 		},
 	}
-	if err := txpool.ValidateTransactionWithState(tx, pool.signer, opts); err != nil {
+}
+
+// validateTx checks whether a transaction is valid according to the consensus
+// rules and adheres to some heuristic limits of the local node (price and size).
+func (pool *LegacyPool) validateTx(tx *types.Transaction, local bool) error {
+	if err := txpool.ValidateTransactionWithState(tx, pool.signer, pool.stateValidationOptions()); err != nil {
 		return err
 	}
 	return pool.validateAuth(tx)
@@ -916,7 +924,6 @@ func (pool *LegacyPool) add(tx *types.Transaction, local bool) (replaced bool, e
 			pendingDiscardMeter.Mark(1)
 			return false, err
 		}
-
 		// New transaction is better, replace old one
 		if old != nil {
 			pool.all.Remove(old.Hash())
@@ -991,7 +998,6 @@ func (pool *LegacyPool) enqueueTx(hash common.Hash, tx *types.Transaction, local
 		queuedDiscardMeter.Mark(1)
 		return false, err
 	}
-
 	// Discard any previous transaction and mark this
 	if old != nil {
 		pool.all.Remove(old.Hash())
@@ -1609,7 +1615,7 @@ func (pool *LegacyPool) promoteExecutables(accounts []common.Address) []*types.T
 			pool.all.Remove(hash)
 		}
 		log.Trace("Removed old queued transactions", "count", len(forwards))
-		// Drop all transactions that are too costly (low balance or out of gas)
+		// Drop queued txs that fail per-tx affordability or block gas checks.
 		drops, _ := list.Filter(pool.chainconfig.IsApplepie(pool.currentHead.Load().Number), pool.currentState, pool.currentState.GetBalance(addr), gasLimit)
 		for _, tx := range drops {
 			hash := tx.Hash()
@@ -1618,16 +1624,52 @@ func (pool *LegacyPool) promoteExecutables(accounts []common.Address) []*types.T
 		log.Trace("Removed unpayable queued transactions", "count", len(drops))
 		queuedNofundsMeter.Mark(int64(len(drops)))
 
-		// Gather all executable transactions and promote them
+		// Gather all executable transactions and promote them. Fee-delegated txs
+		// are re-checked against the shared fee payer's cumulative budget; on the
+		// first failure that tx and its nonce tail are unpromotable, so they are
+		// dropped from the pool (Drop & Forget). Plain txs keep the per-tx behaviour.
 		readies := list.Ready(pool.pendingNonces.get(addr))
-		for _, tx := range readies {
-			hash := tx.Hash()
-			if pool.promoteTx(addr, hash, tx) {
+		before := len(promoted) // promoted is the outer-loop accumulator
+		var (
+			rejectedAtPromote types.Transactions
+			fdOpts            *txpool.ValidationOptionsWithState
+		)
+		for i, tx := range readies {
+			if tx.Type() == types.FeeDelegateDynamicFeeTxType {
+				if fdOpts == nil {
+					fdOpts = pool.stateValidationOptions()
+				}
+				if err := txpool.ValidateTransactionWithState(tx, pool.signer, fdOpts); err != nil {
+					rejectedAtPromote = append(rejectedAtPromote, readies[i:]...)
+					break
+				}
+			}
+			if pool.promoteTx(addr, tx.Hash(), tx) {
 				promoted = append(promoted, tx)
+			} else {
+				// promoteTx already removed tx from the global indexes. Ready removed the
+				// whole prefix from the queue, so only the unprocessed nonce tail needs cleanup.
+				rejectedAtPromote = append(rejectedAtPromote, readies[i+1:]...)
+				break
 			}
 		}
-		log.Trace("Promoted queued transactions", "count", len(promoted))
+		log.Trace("Promoted queued transactions", "count", len(promoted)-before)
 		queuedGauge.Dec(int64(len(readies)))
+
+		// Drop the rejected tail. list.Ready already removed them from the queue
+		// list and reversed their cost accounting (subCosts), and queuedGauge was
+		// decremented above via len(readies); only the global lookup, priced index,
+		// and local gauge remain to clean up.
+		for _, tx := range rejectedAtPromote {
+			pool.all.Remove(tx.Hash())
+		}
+		if n := len(rejectedAtPromote); n > 0 {
+			pool.priced.Removed(n)
+			queuedNofundsMeter.Mark(int64(n))
+			if pool.locals.contains(addr) {
+				localGauge.Dec(int64(n))
+			}
+		}
 
 		// Drop all transactions over the allowed limit
 		var caps types.Transactions
--- a/core/txpool/legacypool/legacypool.go
+++ b/core/txpool/legacypool/legacypool.go
@@ -247,12 +247,13 @@ type LegacyPool struct {
 	locals  *accountSet // Set of local transaction to exempt from eviction rules
 	journal *journal    // Journal of local transaction to back up to disk
 
-	reserver txpool.Reserver              // Address reserver to ensure exclusivity across subpools
-	pending  map[common.Address]*list     // All currently processable transactions
-	queue    map[common.Address]*list     // Queued but non-processable transactions
-	beats    map[common.Address]time.Time // Last heartbeat from each known account
-	all      *lookup                      // All transactions to allow lookups
-	priced   *pricedList                  // All transactions sorted by price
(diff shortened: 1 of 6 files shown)


core/txpool/legacypool/legacypool.go
```diff
@@ -1560,17 +1624,53 @@ func (pool *LegacyPool) promoteExecutables(accounts []common.Address) []*types.T
...
+		for i, tx := range readies {
+			if tx.Type() == types.FeeDelegateDynamicFeeTxType {
+				if fdOpts == nil {
+					fdOpts = pool.stateValidationOptions()
+				}
+				if err := txpool.ValidateTransactionWithState(tx, pool.signer, fdOpts); err != nil {
+					rejectedAtPromote = append(rejectedAtPromote, readies[i:]...)
+					break
+				}
+			}
```
hmlee-wm: Here, isn’t the only necessary check whether the fee payer’s cumulative pendingGas, plus the current transaction’s FeeCost(), exceeds that account’s balance?

Using ValidateTransactionWithState here seems a bit overkill. That function ends up re-running nonce ordering checks and per-transaction balance validation, both of which are already guaranteed during the list.Ready() and add() stages.

hominlee-wemade:
Both failure paths drop the nonce tail more aggressively than upstream

|                     | When nonce 3 fails                                                                                 |
|---------------------|----------------------------------------------------------------------------------------------------|
| Upstream            | Continues the loop — subsequent nonces are attempted, gaps cleaned up later by demoteUnexecutables |
| Current (line 1643) | Fee-delegated validation failure drops nonce 3 and the entire tail                                 |
| Current (line 1652) | promoteTx failure drops the entire tail                                                            |

In both cases the tail drop is too aggressive. Each nonce in readies can have a different fee payer or different cost, so a failure at nonce 3 says nothing about whether nonce 4 is affordable.
Suggest aligning with upstream: on failure, clean up only the failing tx and continue the loop. Any nonce gap left behind will be handled by demoteUnexecutables.

core/txpool/legacypool/legacypool.go
```diff
@@ -1560,17 +1624,53 @@ func (pool *LegacyPool) promoteExecutables(accounts []common.Address) []*types.T
 		log.Trace("Removed unpayable queued transactions", "count", len(drops))
 		queuedNofundsMeter.Mark(int64(len(drops)))
 
-		// Gather all executable transactions and promote them
+		// Gather all executable transactions and promote them. Fee-delegated txs
+		// are re-checked against the shared fee payer's cumulative budget; on the
+		// first failure that tx and its nonce tail are unpromotable, so they are
+		// dropped from the pool (Drop & Forget). Plain txs keep the per-tx behaviour.
```
eomti-wm: In the pre-modified code and the upstream go-ethereum implementation, there seems to be no cumulative validation logic during transaction promotion—not only for fee-delegated transactions but also for plain transactions. 

Therefore, we should reconsider the necessity of adding this cumulative validation logic.

core/txpool/legacypool/legacypool.go
```diff
@@ -981,20 +1035,30 @@ func (pool *LegacyPool) journalTx(from common.Address, tx *types.Transaction) {
...
+		pool.pending[addr] = newPendingList(pool.addPendingGas, pool.subPendingGas)
 	}
 	list := pool.pending[addr]
 
-	inserted, old := list.Add(tx, pool.config.PriceBump)
-	if !inserted {
-		// An older transaction was better, discard this
+	old, err := list.Add(tx, pool.config.PriceBump)
+	if err != nil {
+		log.Error("Promoting invalid queued transaction", "hash", tx.Hash(), "err", err)
```
eomti-wm: It appears that this log was introduced to consume the newly returned error from list.Add, which previously did not return an error.

However, since a transaction failing to promote is a normal occurrence under network competition rather than a node malfunction, using log.Error seems inappropriate. Downgrading it to log.Trace would be more suitable.

core/txpool/legacypool/legacypool.go
colinkim: Upon reviewing the code against upstream, I agree that using ValidateTransactionWithState in the promotion path was broader than necessary.

I removed the promotion-time validation path entirely and aligned promotion behavior with upstream.

Modified in 1d8d36f.

core/txpool/legacypool/legacypool.go
colinkim: fixed in fdfa9f8b3

core/txpool/legacypool/legacypool.go
colinkim: Thank you. After reviewing this against upstream, I agree. Running cumulative validation during promotion changes the semantics of the existing pool and makes failure paths drop the nonce tail more aggressively than upstream.

Therefore, I will revert that part and keep cumulative fee-payer validation limited to tx submission, matching the upstream behavior.

fixed in 1d8d36ffa

core/txpool/legacypool/legacypool_test.go
hmlee-wm, line 1: A few fee-delegation test scenarios that don’t seem to be covered yet, for example:

1. pendingGas on replacement — replacing a pending tx (e.g. FD→plain, plain→FD, FD payerA→FD payerB) and verifying pendingGas updates for both the old and new payer. TestListFeeDelegatedCounter does replace an FD tx, but it runs on a bare list without the pendingGas callbacks wired in, so this path isn’t exercised at the pool level.
2. pendingGas across demote → requeue → repromote — demoteUnexecutables can move a pending FD tx back to queue on a nonce gap. It would be good to ensure pendingGas is decremented on demotion and correctly re-added if the tx is promoted again later. Current tests only cover the drop path, not the requeue/re-promotion cycle.
3. ExistingTx on a pending FD replacement — there is no test that replaces a pending fee-delegated tx (same sender + nonce), so the ExistingTx-based split of value/gas across accounts is not exercised for that path.

0xmhha:
**PR description no longer matches the committed code — "Promotion and demotion" section is stale**

The **Promotion and demotion** section of the description states:

- "Revalidate each ready fee-delegated transaction before promoting it from queue to pending."
- "Account for transactions promoted earlier in the same pass when checking a shared fee payer."
- "On the first promotion failure, drop the rejected transaction and its dependent nonce tail."

None of these are in the current code. Commit `1d8d36ffa` (*revert(txpool): remove promotion-time fee-payer validation*) removed the promotion-time revalidation and deleted its test (`TestFeeDelegationCumulativeGasOnPromotion`). `promoteExecutables` now promotes ready txs via `promoteTx` with no cumulative fee-payer check.

As a result, the third scenario listed in the description ("Disabling it allowed ..." item 3) — *"Queued transactions could bypass the pending-only accounting and become collectively unaffordable when promoted together"* — is **not** addressed by the current code:

- Queue lists are created with `newList(false)` and are not wired into `pendingGas`, so queued FD txs sharing a fee payer each pass the submission-time cumulative check in isolation (the index reads ~0 for them).
- `promoteExecutables` has no cumulative check, so they can all be promoted in a single reorg pass, driving `pendingGas[payer]` above the fee payer's balance.
- `Filter` (used by both promote and demote) only checks each FD tx's `FeeCost()` against the raw `stateDB.GetBalance(feePayer)` per-tx, never cumulatively — so this over-commitment is not corrected afterwards either.

Please either:
- **(a)** if the revert was intentional, update/remove the "Promotion and demotion" section and adjust that third scenario to reflect that cumulative protection is enforced only at submission; or
- **(b)** if the revert was accidental, restore the promotion-path cumulative validation (and `TestFeeDelegationCumulativeGasOnPromotion`).

Note: build and `TestFeeDelegation*` / `TestPendingGasZeroAccounting` / `TestListFeeDelegatedCounter` all pass; this is a description-vs-code consistency issue, not a build break.


I updated the PR description to match the current code

core/txpool/legacypool/legacypool_test.go
colinkim: fixed in d78ab3e48

core/txpool/legacypool/legacypool_test.go
```diff
@@ -3137,3 +3143,532 @@ func BenchmarkMultiAccountBatchInsert(b *testing.B) {
...
+		got = g.ToBig()
+	}
+	pool.mu.RUnlock()
+
+	if got.Cmp(want) != 0 {
+		t.Fatalf("pendingGas[%s] = %v, want %v", payer.Hex(), got, want)
+	}
+}
+
+func TestFeeDelegationPendingGasReplacementAccounting(t *testing.T) {
```
hmlee-wm, line 3269: This test only asserts the fee-payer side (`expectPendingGas`) after each replacement step. The sender-side accounting (`pool.pending[sender].totalcost`) is never checked per step — it's only implicitly covered once, at the very end, via `validateFeeDelegationAccounting`.

Per `list.go`, sender `totalcost` should track `tx.Value()` for FD txs and the full `tx.Cost()` for plain txs. Across this test's four replacements (FD→plain→FD→FD), that value should flip between `value` and the plain tx's full cost each time. A bug that transiently mis-tracks `totalcost` on an intermediate step (e.g. during the plain→FD swap) but happens to cancel out by the final replacement would pass today, since only the end state is verified.

Suggest adding a sender-side counterpart to `expectPendingGas` and calling it after each step, checking `pool.pending[sender].totalcost` against:
- after `fdA` (FD, payerA): `value`
- after `plain`: `plain.Cost()`
- after `fdB` (FD, payerB): `value`
- after `fdA2` (FD, payerA): `value`

colinkim: fixed in 8734d94a6

core/txpool/legacypool/list.go
```diff
@@ -296,17 +308,23 @@ func (l *list) Contains(nonce uint64) bool {
...
-// transaction was accepted, and if yes, any previous transaction it replaced.
+// Add tries to insert a new transaction into the list, returning any previous
+// transaction it replaced, or an error if the transaction was rejected.
 //
-// If the new transaction is accepted into the list, the lists' cost and gas
-// thresholds are also potentially updated.
-func (l *list) Add(tx *types.Transaction, priceBump uint64) (bool, *types.Transaction) {
+// If the new transaction is accepted into the list, the list's cost and gas
+// thresholds and expenditure accounting are also potentially updated.
+func (l *list) Add(tx *types.Transaction, priceBump uint64) (*types.Transaction, error) {
```
hmlee-wm: `Add()` was changed from `(bool, *types.Transaction)` to `(*types.Transaction, error)` to let it distinguish between multiple rejection reasons (`ErrReplaceUnderpriced` vs. `ErrInvalidFeePayer`). Now that the `nil` `FeePayer` case is fully handled by `validateTx` before a tx ever reaches the list, `Add()` only ever returns one error: `ErrReplaceUnderpriced`.

With just one possible failure reason, there's no longer a need to carry a generic `error` here. Please revert `Add()` back to its original `(bool, *types.Transaction)` signature, and update the call sites to match.

core/txpool/legacypool/list.go
```diff
@@ -322,35 +340,32 @@ func (l *list) Add(tx *types.Transaction, priceBump uint64) (bool, *types.Transa
...
 		}
 		// Old is being replaced, subtract old cost
-		l.subTotalCost([]*types.Transaction{old})
+		l.subCosts([]*types.Transaction{old})
 	}
-	// Add new tx cost to totalcost
-	cost, overflow := uint256.FromBig(tx.Cost())
-	if overflow {
-		return false, nil
+	if err := l.addCost(tx); err != nil {
```
hmlee-wm: Every branch here returns `nil` — the `nil` `FeePayer` case that used to return `ErrInvalidFeePayer` is now handled upstream in `validateTx` before a tx reaches the list, so this function can never actually fail.

Please drop the `error` return from this function's signature so it reflects reality, and update the call site in `Add()` to a plain `l.addCost(tx)` instead of `if err := l.addCost(tx); err != nil { ... }`.

core/txpool/legacypool/list.go
colinkim: fixe in 6ea861da2

core/txpool/legacypool/list.go
colinkim: fixed in 6ea861da2

PR #199, opened 2026-06-25
ci: replace echo SSH key with webfactory/ssh-agent

hominlee-wemade:
Replace writing SSH private key to disk via echo with webfactory/ssh-agent action, which loads the key into memory only and never writes it to the filesystem.


hominlee-wemade:
Closing this PR as the dev-test-work workflow is no longer needed. Will be replaced with a PR that removes the workflow entirely.

